PR #1, opened 2020-11-10
Elementd harness

bonomat:
Add elementsd-harness for project droplet.

The docker container image is part of the cargo build process.


src/image.rs
```diff
@@ -0,0 +1,286 @@
...
+    pub accept_non_std_txn: Option<bool>,
+    pub rest: bool,
+    pub validatepegin: bool,
+}
+
+impl Default for ElementsCoreImageArgs {
+    fn default() -> Self {
+        ElementsCoreImageArgs {
+            server: true,
+            network: Network::Regtest,
```
thomaseizinger, line 118: The networks are named differently for elements, no?

build.rs
```diff
@@ -0,0 +1,25 @@
+use std::process::Command;
+
+fn main() {
```
thomaseizinger, line 3: This build script should probably only run if the dockerfile changes: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed

src/elementd_rpc.rs
```diff
@@ -0,0 +1,75 @@
...
+            ))
+            .await?;
+
+        Ok(blockchain_info)
+    }
+
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
thomaseizinger, line 44: What is the point of an enum with a single variant?

src/json_rpc.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(untagged)]
+pub enum Response<T> {
+    Success { result: T },
+    Error { error: JsonRpcError },
+    RpcError(JsonRpcError),
+}
```
thomaseizinger, line 88: This way of modelling the response does not produce very good error messages. See https://github.com/comit-network/comit-rs/pull/3262.

D4nte:
Looks like this repo should be called `elements-harness-rs`

src/image.rs
bonomat: True, slightly different and regtest can be an arbitrary string.

build.rs
bonomat: Cool, I almost wrote this myself 😅

bonomat:
@D4nte 
> Looks like this repo should be called elements-harness-rs

No, the repository contains Project Droplet specific content. The elements-harness functionality is just the start to simplify testing.

elements-harness/build.rs
```diff
@@ -16,10 +15,10 @@ fn main() {
             let stderr = String::from_utf8(output.stderr)
                 .unwrap_or_else(|_| String::from("Could not decode stderr"));
```
thomaseizinger, line 16: That looks like you misunderstood what `unwrap_or_else` does? You are assigning the string "Could not decode stderr" to `stderr` here in case the `from_utf8` fails. Might be better to panic here, this is just a build-script anyway!

Sorry I missed this on the first pass!

thomaseizinger (approved):

elements-harness/build.rs
bonomat: No worries. Fixed now.

PR #2, opened 2020-11-10
Replace custom json rpc with Thomas' json-rpc

bonomat:
Use Thomas' rpc client instead of custom rpc client


elements-harness/src/image.rs
```diff
@@ -143,7 +144,7 @@ impl IntoIterator for ElementsCoreImageArgs {
 
         match self.network {
             Network::Testnet => args.push("-testnet".to_string()),
-            Network::Regtest => args.push("-regtest".to_string()),
```
bonomat, line 146: I honestly don't know what the difference between `regtest` and a custom chain is. I followed the instructions from here: https://docs.blockstream.com/liquid/node_setup.html

elements-harness/src/image.rs
```diff
@@ -183,6 +185,11 @@ impl IntoIterator for ElementsCoreImageArgs {
 
         args.push("-debug".into());
 
+        args.push("-con_elementsmode=true".into());
```
bonomat, line 188: Not needed, will default to true.
```suggestion
```

elements-harness/src/elementd_rpc.rs
```diff
@@ -1,65 +1,121 @@
...
 
-            (Client::new(blockchain.node_url.clone()), blockchain)
+        let blockchain_info: BlockchainInfo = client.getblockchaininfo().await.unwrap();
+        let network = blockchain_info.chain;
+
+        assert_eq!(network, "elementsregtest")
+    }
+
+    #[tokio::test]
+    async fn rpc_sanity_tests() {
```
luckysori, line 85: :thinking: I'm not a big fan of tests that test too much, but I guess you want to avoid spinning up a container for every test. I would rather have individual tests and take the test speed hit though.

luckysori (approved):
Using this new library looks great! Merge asap :stuck_out_tongue:

elements-harness/src/elementd_rpc.rs
bonomat: Me neither. I'm not sure if we should keep this test at all... I rather wanted to play around with the library. 

Shall I remove the test and we only add tests for functions as needed?

luckysori: That sounds good!

bonomat: ... remove ...

PR #6, opened 2020-11-17
Make better use of the type system

thomaseizinger:
See commit messages.

### DIFF
--- a/elements-fun/src/confidential.rs
+++ b/elements-fun/src/confidential.rs
@@ -111,7 +111,7 @@ macro_rules! impl_confidential_commitment {
         }
 
         impl Decodable for $name {
-            fn consensus_decode<D: io::Read>(mut d: D) -> Result<$name, encode::Error> {
+            fn consensus_decode<D: io::BufRead>(mut d: D) -> Result<$name, encode::Error> {
                 let prefix = u8::consensus_decode(&mut d)?;
                 match prefix {
                     0 => Ok($name::Null),
--- a/elements-fun/src/confidential.rs
+++ b/elements-fun/src/confidential.rs
@@ -23,109 +23,66 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{fmt, io};
 
 use crate::encode::{self, Decodable, Encodable};
-use crate::issuance::AssetId;
 
 // Helper macro to implement various things for the various confidential
 // commitment types
 macro_rules! impl_confidential_commitment {
-    ($name:ident, $inner:ty, $prefixA:expr, $prefixB:expr) => {
-        impl_confidential_commitment!($name, $inner, $prefixA, $prefixB, |x| x);
-    };
-    ($name:ident, $inner:ty, $prefixA:expr, $prefixB:expr, $explicit_fn:expr) => {
+    ($name:ident, $prefixA:expr, $prefixB:expr) => {
         impl $name {
-            /// Create from commitment.
-            pub fn from_commitment(bytes: &[u8]) -> Result<$name, encode::Error> {
-                if bytes.len() != 33 {
+            pub const fn is_valid_prefix(tag: u8) -> bool {
+                tag == $prefixA || tag == $prefixB
+            }
+
+            pub fn new(tag: u8, commitment: &[u8]) -> Result<Self, encode::Error> {
+                if commitment.len() != 32 {
                     return Err(encode::Error::ParseFailed(
-                        "commitments must be 33 bytes long",
+                        "commitments must be 32 bytes long",
                     ));
                 }
-                let prefix = bytes[0];
-                if prefix != $prefixA && prefix != $prefixB {
-                    return Err(encode::Error::InvalidConfidentialPrefix(prefix));
+
+                if !Self::is_valid_prefix(tag) {
+                    return Err(encode::Error::InvalidConfidentialPrefix(tag));
                 }
-                let mut c = [0; 32];
-                c.copy_from_slice(&bytes[1..]);
-                Ok($name::Confidential(prefix, c))
+                let mut bytes = [0u8; 33];
+                bytes[0] = tag;
+                bytes[1..].copy_from_slice(&commitment);
+
+                Ok(Self(bytes))
             }
 
-            /// Check if the object is null.
-            pub fn is_null(&self) -> bool {
-                matches!(*self, $name::Null)
+            pub fn from_slice(bytes: &[u8]) -> Result<$name, encode::Error> {
+                Self::new(bytes[0], &bytes[1..])
             }
 
-            /// Check if the object is explicit.
-            pub fn is_explicit(&self) -> bool {
-                matches!(*self, $name::Explicit(_))
+            pub fn commitment(&self) -> [u8; 33] {
+                self.0
             }
 
-            /// Check if the object is confidential.
-            pub fn is_confidential(&self) -> bool {
-                matches!(*self, $name::Explicit(_))
-            }
-
-            /// Returns the explicit inner value.
-            /// Returns [None] if [is_explicit] returns false.
-            pub fn explicit(&self) -> Option<$inner> {
-                match *self {
-                    $name::Explicit(i) => Some(i),
-                    _ => None,
-                }
-            }
-
-            /// Returns the confidential commitment in case of a confidential value.
-            /// Returns [None] if [is_confidential] returns false.
-            pub fn commitment(&self) -> Option<[u8; 33]> {
-                match *self {
-                    $name::Confidential(p, c) => {
-                        let mut res = [0; 33];
-                        res[0] = p;
-                        res[1..].copy_from_slice(&c[..]);
-                        Some(res)
-                    }
-                    _ => None,
-                }
-            }
-        }
-
-        impl Default for $name {
-            fn default() -> Self {
-                $name::Null
+            pub fn encoded_length(&self) -> usize {
+                33
             }
         }
 
         impl Encodable for $name {
             fn consensus_encode<S: io::Write>(&self, mut s: S) -> Result<usize, encode::Error> {
-                match *self {
-                    $name::Null => 0u8.consensus_encode(s),
-                    $name::Explicit(n) => {
-                        1u8.consensus_encode(&mut s)?;
-                        // Apply $explicit_fn to allow `Value` to swap the amount bytes
-                        Ok(1 + $explicit_fn(n).consensus_encode(&mut s)?)
-                    }
-                    $name::Confidential(prefix, bytes) => {
-                        Ok(prefix.consensus_encode(&mut s)? + bytes.consensus_encode(&mut s)?)
-                    }
-                }
+                self.0.consensus_encode(&mut s)
             }
         }
 
         impl Decodable for $name {
             fn consensus_decode<D: io::BufRead>(mut d: D) -> Result<$name, encode::Error> {
-                let prefix = u8::consensus_decode(&mut d)?;
-                match prefix {
-                    0 => Ok($name::Null),
-                    1 => {
-                        // Apply $explicit_fn to allow `Value` to swap the amount bytes
-                        let explicit = $explicit_fn(Decodable::consensus_decode(&mut d)?);
-                        Ok($name::Explicit(explicit))
-                    }
-                    p if p == $prefixA || p == $prefixB => {
-                        let commitment = <[u8; 32]>::consensus_decode(&mut d)?;
-                        Ok($name::Confidential(p, commitment))
-                    }
-                    p => return Err(encode::Error::InvalidConfidentialPrefix(p)),
+                let bytes = <[u8; 33]>::consensus_decode(&mut d)?;
+
+                Ok(Self::new(bytes[0], &bytes[1..])?)
+            }
+        }
+
+        impl fmt::Display for $name {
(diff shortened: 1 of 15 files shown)


elements-fun/src/confidential.rs
```diff
@@ -23,109 +23,66 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
-                if bytes.len() != 33 {
+            pub const fn is_valid_prefix(tag: u8) -> bool {
+                tag == $prefixA || tag == $prefixB
+            }
+
+            pub fn new(tag: u8, commitment: &[u8]) -> Result<Self, encode::Error> {
+                if commitment.len() != 32 {
                     return Err(encode::Error::ParseFailed(
-                        "commitments must be 33 bytes long",
+                        "commitments must be 32 bytes long",
```
luckysori, line 39: :question: I know we talked about this yesterday, but the tag is part of the commitment, right? It's a nitpick, but it seems incorrect to say that commitments must be 32 bytes long. I guess the argument `commitment` is actually the commitment's x-coordinate. It may not be worth it to disambiguate though.

elements-fun/src/confidential.rs
```diff
@@ -176,26 +120,15 @@ macro_rules! impl_confidential_commitment {
...
-                                None => Err(A::Error::custom("missing commitment")),
-                            },
-                            p => {
-                                return Err(A::Error::custom(format!(
-                                    "invalid commitment, invalid prefix: 0x{:02x}",
-                                    p
-                                )))
-                            }
+                        if prefix != $prefixA && prefix != $prefixB {
+                            return Err(A::Error::custom("missing commitment"));
```
luckysori, line 135: :question: Do you mean "missing prefix"?

elements-fun/src/confidential.rs
```diff
@@ -176,26 +120,15 @@ macro_rules! impl_confidential_commitment {
                             return Err(A::Error::custom("missing prefix"));
```
luckysori, line 131: :question: Is this more like an "empty commitment" error?

elements-fun/src/wally.rs
```diff
@@ -100,11 +101,11 @@ pub fn bip39_mnemonic_to_seed(mnemonic: &str, passphrase: &str) -> Option<[u8; B
 
 /// Calculate the signature hash for a specific index of
 /// an Elements transaction.
-pub fn tx_get_elements_signature_hash(
+pub fn tx_get_elements_signature_hash<V: Encodable>(
     tx: &crate::Transaction,
     index: usize,
     script_code: &bitcoin::Script,
-    value: &crate::confidential::Value,
+    value: &V, // TODO: Is this really the best way to do it?
```
luckysori, line 108: :question: Should we introduce a `ValueCommitment` type?

elements-fun/src/wally.rs
```diff
@@ -353,7 +352,11 @@ pub fn asset_final_vbf(
     final_vbf
 }
 
-pub fn asset_value_commitment(value: u64, vbf: [u8; 32], generator: Asset) -> Value {
+pub fn asset_value_commitment(
+    value: u64,
+    vbf: [u8; 32],
+    generator: AssetCommitment,
+) -> ValueCommitment {
```
luckysori, line 361: :thinking: It's interesting that `libwally` calls this `asset_value_commitment` but we choose `ValueCommitment` as the type.

luckysori: We could change the name of the function.

elements-fun/src/wally.rs
luckysori: It actually exists. Should we use it here?

elements-fun/src/transaction.rs
```diff
@@ -23,34 +23,196 @@ use bitcoin::blockdata::script::{Instruction, Script};
...
 
-use crate::confidential;
-use crate::encode::{self, Decodable, Encodable};
+use crate::confidential::AssetCommitment;
+use crate::confidential::NonceCommitment;
+use crate::confidential::ValueCommitment;
+use crate::encode::{self, Decodable, Encodable, Error};
 use crate::issuance::AssetId;
 
+// TODO: Is `Default` really useful here?
```
luckysori, line 35: :question: Why not? A null asset issuance is what I would expect when I need the rest of a transaction to be default.

elements-fun/src/transaction.rs
```diff
@@ -23,34 +23,196 @@ use bitcoin::blockdata::script::{Instruction, Script};
...
-    pub amount: confidential::Value,
+    pub amount: ExplicitValue,
     /// Amount of inflation keys to issue
-    pub inflation_keys: confidential::Value,
+    pub inflation_keys: ExplicitValue,
+}
+
+impl ExplicitAssetIssuance {
+    pub fn encoded_length(&self) -> usize {
+        32 + 32 + self.amount.encoded_length() + self.inflation_keys.encoded_length()
```
luckysori, line 84: :question: Should we use a `SECRET_KEY_LENGTH` sort of constant (that may already exist) instead of magic numbers?

elements-fun/src/transaction.rs
```diff
@@ -23,34 +23,196 @@ use bitcoin::blockdata::script::{Instruction, Script};
...
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(crate = "serde_crate")
+)]
+pub struct NullAssetIssuance {
+    /// Zero for a new asset issuance; otherwise a blinding factor for the input
+    pub asset_blinding_nonce: [u8; 32],
+    /// Freeform entropy field
+    pub asset_entropy: [u8; 32],
+}
```
luckysori, line 99: :question: Can you explain why a null asset issuance has any fields at all? Doesn't make sense to me given the name.

elements-fun/src/transaction.rs
```diff
@@ -23,34 +23,196 @@ use bitcoin::blockdata::script::{Instruction, Script};
...
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(crate = "serde_crate")
+)]
+pub struct ConfidentialAssetIssuance {
+    /// Zero for a new asset issuance; otherwise a blinding factor for the input
+    pub asset_blinding_nonce: [u8; 32],
```
luckysori, line 115: :question: Have you thought about using `SecretKey` for some of these fields? Would it make sense?

luckysori (commented):
Half-review

elements-fun/src/confidential.rs
thomaseizinger: Fair, it just weird because here I am only deserializing the last 32 bytes :D

I guess this can be an implementation detail though and the error message should be better!

elements-fun/src/confidential.rs
thomaseizinger: Probably, this was copy pasted :D

elements-fun/src/wally.rs
thomaseizinger: Well, you can use this function for confidential and non-confidential values. The new type system doensn't offer anything here.

elements-fun/src/transaction.rs
thomaseizinger: Well, the data is still there I think, like I need to deserialize it. Will have to do some testing to see if this can be improved!

elements-fun/src/transaction.rs
thomaseizinger: Haven't yet thought about it, wanted to get the big refactoring in first1

elements-fun/src/transaction.rs
luckysori: It's just weird that there has to be any data at all if the transaction does not issue an asset. Unless this data is unrelated to asset issuances?

elements-fun/src/transaction.rs
```diff
@@ -170,14 +332,14 @@ impl TxInWitness {
...
-#[derive(Copy, Clone, Default, PartialEq, Eq, Debug, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
 pub struct PeginData<'tx> {
     /// Reference to the pegin output on the mainchain
     pub outpoint: bitcoin::OutPoint,
     /// The value, in satoshis, of the pegin
     pub value: u64,
     /// Asset type being pegged in
-    pub asset: confidential::Asset,
+    pub asset: AssetId,
```
luckysori, line 345: :question: Does this mean that this struct is always explicit?

elements-fun/src/transaction.rs
```diff
@@ -384,20 +544,167 @@ pub struct PegoutData<'txo> {
...
+    Confidential(ConfidentialTxOut),
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(crate = "serde_crate")
+)]
+pub struct NullTxOut {
```
luckysori, line 570: :question: When does elements use null txouts? Just curious.

elements-fun/src/transaction.rs
```diff
@@ -330,16 +547,167 @@ pub struct PegoutData<'txo> {
...
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(crate = "serde_crate", transparent)
+)]
+pub struct ExplicitValue(pub u64);
```
luckysori, line 639: :question: Should we be using `bitcoin::Amount` under the hood? Is that a horrible idea?

elements-fun/src/transaction.rs
```diff
@@ -348,32 +716,201 @@ pub struct TxOut {
...
+
+        Ok(UnblindedTxOut {
+            asset: AssetId::from_slice(&unblinded_asset).unwrap(),
+            original_asset: self.asset,
+            asset_blinding_factor: abf,
+            value_blinding_factor: vbf,
+            value: value_out,
+        })
+    }
+}
```
luckysori, line 751: :tada: This makes it all worth it! Awesome work.

elements-fun/src/transaction.rs
```diff
@@ -348,32 +716,201 @@ pub struct TxOut {
...
+
+pub struct UnblindedTxOut {
+    pub asset: AssetId,
+    pub value: u64,
+    pub original_asset: AssetCommitment, // TODO: should we actually return this?
```
luckysori, line 756: Nah.

elements-fun/src/transaction.rs
```diff
@@ -469,37 +1006,37 @@ impl TxOut {
...
-                    let has_nonzero_range = self.witness.rangeproof[0] & 64 == 64;
-                    let has_min = self.witness.rangeproof[0] & 32 == 32;
+                    let has_nonzero_range = witness.rangeproof[0] & 64 == 64;
+                    let has_min = witness.rangeproof[0] & 32 == 32;
 
                     if !has_min {
                         min_value
                     } else if has_nonzero_range {
-                        bitcoin::consensus::deserialize::<u64>(&self.witness.rangeproof[2..10])
+                        bitcoin::consensus::deserialize::<u64>(&witness.rangeproof[2..10])
```
luckysori, line 1035: :thinking: Imagine if we had a rangeproof struct.

elements-fun/src/transaction.rs
```diff
@@ -1558,32 +2093,33 @@ mod tests {
...
+        //             9,
+        //             [
+        //                 0x81, 0x65, 0x4e, 0xb5, 0xcc, 0xd9, 0x92, 0x7b, 0x8b, 0xea, 0x94, 0x99,
+        //                 0x7d, 0xce, 0x4a, 0xe8, 0x5b, 0x3d, 0x95, 0xa2, 0x07, 0x00, 0x38, 0x4f,
+        //                 0x0b, 0x8c, 0x1f, 0xe9, 0x95, 0x18, 0x06, 0x38
+        //             ],
+        //         ),
+        //         inflation_keys: confidential::Value::Null,
+        //     }
+        // );
```
luckysori, line 2111: :wrench: I guess you meant to remove it.

luckysori (commented):
This looks really good to me! I would consider upstreaming this. The worst that can happen is that they say no and we just continue with our own version of it.

elements-fun/src/transaction.rs
thomaseizinger: Yes from my understanding yes!

elements-fun/src/transaction.rs
thomaseizinger: I don't fully understand this yet :D

elements-fun/src/transaction.rs
thomaseizinger: Not a horrible idea because all tokens have the same decimal points :)

elements-fun/src/transaction.rs
thomaseizinger: It is all possible :D

elements-fun/src/transaction.rs
thomaseizinger: Ups, no I didn't!

Will put it back!

PR #12, opened 2020-12-01
Init commit create react app with react-app-rewired

bonomat:
# What is this?

This adds a new `waves` folder which is a react app using a rust wasm library combined with webpack.

`my-app` was creating using create-react-app.
However, this would not let us configure webpack (which is the reason we went for create-react-app in the first place).
Hence we use `react-app-rewired` and created a `config-override.js`
The library `wasm_lib` was created using `wasm-pack new`.
It showcases a single method to return hello world and a function to initialize a logger.

# Issues

So far I like the setup, it has one caveat though: if we want to add `native` as a dependency in `package.toml` then a rebuild is not automatically reloaded. File-based import work though.

# How to

Go into `my-app` and type
```bash
yarn install
```
followed by 
```bash
yarn start
```
to start in development mode. You can then make changes in the rust library under `native` and the code should automatically rebuild and reload. 

Enjoy :)


waves/.gitignore
```diff
@@ -0,0 +1,25 @@
...
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+.eslintcache
```
thomaseizinger, line 25: Missing newline.

waves/config-overrides.js
```diff
@@ -0,0 +1,25 @@
...
+
+    config.plugins = (config.plugins || []).concat([
+        new WasmPackPlugin({
+            crateDirectory: path.resolve(__dirname, "src/native/"),
+            outDir: path.resolve(__dirname, "src/native/pkg")
+        }),
+    ]);
+
+    return config;
+}
```
thomaseizinger, line 25: Missing newline.

Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = [ "elements-harness", "swap", "elements-fun" ]
+members = [ "elements-harness", "swap", "elements-fun", "waves/src/native" ]
```
thomaseizinger, line 2: Is `waves` really such a good name considering there is a whole blockchain ecosystem named like that?

waves/src/App.tsx
```diff
@@ -0,0 +1,20 @@
+import React, {useState} from 'react';
+import './App.css';
+
+function App() {
+    import('./native/pkg').then(({ hello }) => {
+        let welcome = hello("World");
+        setWelcome(welcome)
+    })
```
thomaseizinger, line 8: Can we only use async imports with wasm?

What happens if you try to regularly import this at the top with `import wasm from "./native/pkg"`? Will the wasm get inlined into the JS or is it still loaded as a dedicated resource?

Ideally, we can just use regular imports but the wasm binary is loaded as a dedicated resource.

Potentially relevant: https://webpack.js.org/guides/code-splitting/#dynamic-imports

waves/src/native/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+use crate::utils::set_panic_hook;
+use wasm_bindgen::prelude::*;
+
+// When the `wee_alloc` feature is enabled, use `wee_alloc` as the global
+// allocator.
+#[cfg(feature = "wee_alloc")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[wasm_bindgen]
```
thomaseizinger, line 13: Should this maybe be a `wasm_bindgen(start)`? https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-rust-exports/start.html

waves/src/native/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "wasm_lib"
```
thomaseizinger, line 2:
```suggestion
name = "waves-wallet"
```

? :)

waves/src/App.test.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import App from './App';
+
+test('renders learn react link', () => {
+  render(<App />);
+  const linkElement = screen.getByText(/learn react/i);
+  expect(linkElement).toBeInTheDocument();
+});
```
thomaseizinger, line 9: Does that still pass with the changes you made?

waves/src/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  <React.StrictMode>
+    <App />
+  </React.StrictMode>,
+  document.getElementById('root')
+);
+
+// If you want to start measuring performance in your app, pass a function
+// to log results (for example: reportWebVitals(console.log))
+// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
+reportWebVitals();
```
thomaseizinger, line 20: I would be interested in keeping this! Can we pass `console.log` here so that we can see how the usage of `wasm` affects performance? Might also be interesting in regards to the async/sync import story.

thomaseizinger (commented):
Thanks for getting this bootstrapped!

I think we should also add CI scripts right away to make sure we are not accidentally breaking stuff again.

waves/src/App.test.tsx
bonomat: Nope 😅

Cargo.toml
bonomat: Happy to rename it :) Back to Greek god names?

waves/src/App.tsx
bonomat: > Can we only use async imports with wasm?

I tried this at first but it wasm imports need to be async.
webpack 5 should solve this issue but it's not yet the default version in create-react-app.

- https://webpack.js.org/configuration/experiments/
- https://github.com/facebook/create-react-app/issues/9994

waves/src/index.tsx
bonomat: This produces quite some noise in the logs though.  I'll leave it for now to keep the console clean.

bonomat:
superseded by: https://github.com/comit-network/droplet/pull/13

thomaseizinger:
> superseded by: #13

Why not just rebase onto latest master instead of closing?

PR #13, opened 2020-12-02
Init commit create react app with react-app-rewired

bonomat:
supersedes https://github.com/comit-network/droplet/pull/12 do to changes of master

# What is this?

This adds a new `waves` folder which is a react app using a rust wasm library combined with webpack.

`my-app` was creating using create-react-app.
However, this would not let us configure webpack (which is the reason we went for create-react-app in the first place).
Hence we use `react-app-rewired` and created a `config-override.js`
The library `wasm_lib` was created using `wasm-pack new`.
It showcases a single method to return hello world and a function to initialize a logger.

# Issues

So far I like the setup, it has one caveat though: if we want to add `native` as a dependency in `package.toml` then a rebuild is not automatically reloaded. File-based import work though.

# How to

Go into `my-app` and type
```bash
yarn install
```
followed by 
```bash
yarn start
```
to start in development mode. You can then make changes in the rust library under `native` and the code should automatically rebuild and reload. 

Enjoy :)


.github/workflows/ci.yml
```diff
@@ -62,8 +62,21 @@ jobs:
           path: ~/.cargo/registry
           key: rust-${{ matrix.rust_toolchain }}-cargo-registry-directory-${{ hashFiles('Cargo.lock') }}-v1
 
+      - name: Cache node_modules directory
+        uses: actions/cache@v1
+        with:
+          path: waves/node_modules
+          key: node-${{ matrix.rust_toolchain }}-node-modules-directory-${{ hashFiles('waves/yarn.lock') }}-v1
```
thomaseizinger, line 69: Why is the `rust_toolchain` part of the node cache key?

bobtimus/src/main.rs
```diff
@@ -0,0 +1,17 @@
+#![deny(warnings)]
+use warp::Filter;
+
+#[tokio::main]
+async fn main() {
+    let cors = warp::cors().allow_any_origin();
```
thomaseizinger, line 6: We shouldn't need cors if we serve the page from the same port.

waves/src/App.tsx
```diff
@@ -0,0 +1,31 @@
...
+function App() {
+    const [welcome, setWelcome] = useState<String>("Not welcome yet");
+    const [rate, setRate] = useState<String>("No rate received yet");
+
+    import("./wallet-lib/pkg").then(({ hello }) => {
+        let welcome = hello("World");
+        setWelcome(welcome);
+    });
+
+    fetch("http://localhost:3030/rate")
```
thomaseizinger, line 13: In prod, this should simply be:

```suggestion
    fetch("/rate")
```

thomaseizinger: `webpack-dev-server` supports a proxy: https://webpack.js.org/configuration/dev-server/#devserverproxy

Means you can tell it to re-write certain requests and point them against some sub-server then you don't have to poison the prod code with conditionals :)

waves/src/App.tsx
```diff
@@ -0,0 +1,31 @@
...
+import "./App.css";
+
+function App() {
+    const [welcome, setWelcome] = useState<String>("Not welcome yet");
+    const [rate, setRate] = useState<String>("No rate received yet");
+
+    import("./wallet-lib/pkg").then(({ hello }) => {
+        let welcome = hello("World");
+        setWelcome(welcome);
+    });
```
thomaseizinger, line 11: I feel this is not gonna be the endgame of how we can/want to use wasm functionality.

What happens if you call this code twice? Will it load the wasm file twice from the server?
Another interesting thing to explore would be to hide this behind a regular `async` function. `import()` simply returns a `Promise` which means you can await it. As such, you could just expose a `async function hello()` in TS and we wouldn't need to be concerned here with the fact that we need to import the wasm dynamically. That ofc only makes sense if we don't load the WASM again for every function call but this is somehow cached by webpack!

thomaseizinger: If it is not cached, we might want to abstract this away in a custom hook that caches the wasm instance using `useState`, similar to what [this](https://codeburst.io/getting-started-with-react-and-webassembly-using-hooks-441818c91608) blogpost is doing!

Happy to look into this tomorrow together if you want :)

The reason I am pushing for this is because I think it would be good if we have a relatively clear idea of how we are going to consume the WASM code from within React that doesn't cause a lot of boilerplate!

.github/workflows/ci.yml
bonomat: copy pasta :)

waves/src/App.tsx
bonomat: I had a look at this but couldn't get it working at the first attempt so I stopped.

waves/src/App.tsx
bonomat: I agree, this is not optimal. Happy to look into this together :)

bobtimus/src/main.rs
bonomat: resolved due to proxy.

waves/src/wasmProxy.ts
```diff
@@ -0,0 +1,4 @@
+export async function hello(msg: string) {
```
bonomat, line 1: @thomaseizinger : what do you think about this simple approach for now?

PR #24, opened 2020-12-10
Don't store any secret values in plain text

thomaseizinger:


waves/wallet/src/wallet.rs
```diff
@@ -285,14 +349,40 @@ impl Wallet {
...
             .expect("output length aligns with sha256");
 
         SecretKey::from_slice(bk.as_ref()).expect("always a valid secret key")
     }
+
+    /// Derive the encryption key from the wallet's name and password.
+    ///
+    /// # Choice of salt
+    ///
+    /// The salt of HKDF can be public or secret and while it can operate without a salt, it is better to pass a salt value [0].
```
luckysori, line 365: :question: If we're gonna use a salt shouldn't it be random?

waves/wallet/src/wallet.rs
```diff
@@ -249,12 +258,50 @@ pub struct BalanceEntry {
...
 pub struct Wallet {
-    pub name: String,
-    pub encryption_key: [u8; 32],
-    pub secret_key: SecretKey,
+    name: String,
+    encryption_key: [u8; 32],
+    secret_key: SecretKey,
 }
 
+const SECRET_KEY_ENCRYPTION_NONCE: &[u8; 12] = b"SECRET_KEY!!";
```
luckysori, line 266: :sob:

luckysori (approved):
Thanks for documenting your choices! It looks good to me.

waves/wallet/src/wallet.rs
thomaseizinger: If the salt is random, I need to store it on disk next to all the other data. I was thinking about but I wanted to store less data overall not more.

However, if the user deletes their browser storage, things are lost anyway ...

thomaseizinger:
Rebased to fix conflicts.

waves/wallet/src/wallet.rs
thomaseizinger: What do you think about the choice of a static nonce? We could also generate a random number so not every user uses the same one. Thoughts?

thomaseizinger:
Depends on https://github.com/comit-network/droplet/pull/26 to fix build errors.

waves/wallet/src/wallet.rs
luckysori: I was just amused by the string, but it's a good question. I think they ciphertext will always be unique because the plaintext is generated, so it's probably fine.

waves/wallet/src/wallet.rs
luckysori: I don't think it's very dangerous, but two users could have the same encryption key if they use the same name and password. Discovering the encryption key of one could make the other vulnerable, but of course our users are not in the same place so the risk is probably negligible.

thomaseizinger: See my last commit, I am using a random salt now!

thomaseizinger:
@luckysori A final patch that enables the weak `ScryptParams` only in debug mode.

It is really important that no user ever _creates_ a mainnet account with a debug build. Loading one is fine but will take a long time.

cc @bonomat

PR #27, opened 2020-12-11
Change Bobtimus' API to accept an outpoint instead of TxIn and TxOut

thomaseizinger:
Bobtimus can easily query the blockchain for all data it needs, hence
we can reduce bandwidth usage and simplify client implementations.

### DIFF
--- a/bobtimus/src/lib.rs
+++ b/bobtimus/src/lib.rs
@@ -1,13 +1,15 @@
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
 use elements_fun::{
+    bitcoin::Amount,
     secp256k1::{
         rand::{CryptoRng, RngCore},
         SecretKey,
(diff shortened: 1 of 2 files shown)


bobtimus/src/lib.rs
```diff
@@ -84,7 +96,62 @@ impl<R, RS> Bobtimus<R, RS> {
...
+                            })?
+                            .clone();
+
+                        Result::<_, anyhow::Error>::Ok((txin, txin_as_txout, blinding_key))
+                    }
+                },
+            )
+            .collect::<FuturesUnordered<_>>()
+            .try_collect::<Vec<_>>()
+            .await?;
```
thomaseizinger, line 138: I freakin' love how concise this is in Rust!

This executes all futures in parallel and returns us a `Result<Vec<_>, _>`!

luckysori (approved):
Looks really good!

PR #28, opened 2020-12-11
Introduce real bobtimus

### LINKED ISSUE #17: Bobtimus
Bobtimus is in the role of Bob

- rate feed: publish fixed rate every X seconds
- api to implement protocol steps
	- interaction with elementsd
	- user provides an elementsd node with an existing wallet
	- we expect the wallet to remain unlocked forever

luckysori:
Closes #17.

A lot of silly debugging and compiler fighting for something quite simple, but it works.


bobtimus/src/bin/bobtimus.rs
```diff
@@ -0,0 +1,204 @@
...
+        "name": "ticker"
+      }
+    }"#;
+
+    #[derive(Clone)]
+    pub struct RateService(Receiver<Rate>);
+
+    #[async_trait]
+    impl LatestRate for RateService {
+        async fn latest_rate(&mut self) -> anyhow::Result<Rate> {
```
thomaseizinger, line 82: So this implementation will currently block on the next rate before it can proceed whereas ideally, we want it to use the last one we sent to the user.

Instead of having this functionality here, you could just make another struct that - given a stream of rates - always stores the latest value in a local field and makes it available through a getter. (For that you will need spawn a task with a loop into the runtime)

You might want to use a `RwLock` for that instead of a `Mutex` (so that different clients don't block each other).

thomaseizinger (approved):
Good effort actually using the WS API of Kraken! No polling :)

bobtimus/src/bin/bobtimus.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+
+        #[tokio::test]
+        async fn deserialize_ticker_update() {
+            let sample_response = r#"
```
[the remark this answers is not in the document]
luckysori: I hate it so much.

bobtimus/src/bin/bobtimus.rs
```diff
@@ -0,0 +1,204 @@
...
+    use watch::Receiver;
+
+    const KRAKEN_WS_URL: &str = "wss://ws.kraken.com";
+    const SUBSCRIBE_XBT_USDT_TICKER_PAYLOAD: &str = r#"
+    { "event": "subscribe",
+      "pair": [ "XBT/USD" ],
+      "subscription": {
+        "name": "ticker"
+      }
+    }"#;
```
bonomat, line 75: Nitpick: Either rename variable to `...XBT_USD_TICKER` or subscribe to `XBT/USDT` because  
`USD` is the fiat asset while `USDT` is the crypto asset (the stable coin).

```suggestion
    const SUBSCRIBE_XBT_USDT_TICKER_PAYLOAD: &str = r#"
    { "event": "subscribe",
      "pair": [ "XBT/USDT" ],
      "subscription": {
        "name": "ticker"
      }
    }"#;
```

bonomat (approved):
Awesome work and welcome Bobtimus :)

bobtimus/src/bin/bobtimus.rs
luckysori: Good point! I think I'll go for the rename, because I tried the pair `XBT/USDT` and the subscription produced very few updates over time.

bobtimus/src/amounts.rs
```diff
@@ -8,7 +8,7 @@ use std::{convert::TryFrom, fmt::Debug};
 ///
 /// - The `ask` represents the minimum price for which we are willing to sell 1 L-BTC.
 /// - The `bid` represents the maximum price we are willing pay for 1 L-BTC.
-#[derive(Debug, Clone, Copy, Serialize)]
+#[derive(Default, Debug, Clone, Copy, Serialize)]
```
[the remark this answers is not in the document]
luckysori: I was not too happy with this either. It stems from the need to start a `tokio::sync::watch::channel` with an initial value. A zero rate is the least dangerous value to put in the channel before we connect to Kraken, but it's not necessarily the best use of the `Default` for the `Rate` type.

bobtimus/src/bin/bobtimus.rs
luckysori: I've solved it (I think) by implementing `Future` on `RateService` and always looking for the last value in the stream, updating an internal state and only if the future is pending actually returning the stored state. Take a look: https://github.com/comit-network/droplet/pull/28/commits/d1f13225d7ee40209c05cb649a5a191e07f0c48b.

bobtimus/src/amounts.rs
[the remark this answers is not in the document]
luckysori: I've replaced it with a `ZERO` constant that is only used internally and is never actually exposed to consumers of the stream or `LatestRate`.

bobtimus/src/bin/bobtimus.rs
thomaseizinger: Interesting solution! :)

bobtimus/src/bin/bobtimus.rs
```diff
@@ -200,5 +209,27 @@ mod kraken {
...
+
+            let mut service = RateService {
+                receiver: read,
+                latest_rate: Rate::ZERO,
+            };
+
+            let rate = service.latest_rate().await.unwrap();
+            assert_eq!(rate, latest_rate);
+
+            let _rate = service.latest_rate().await.unwrap();
```
thomaseizinger, line 231: You will have to use the return value of this to make the test meaningful no?

bobtimus/src/bin/fake_bobtimus.rs
```diff
@@ -26,26 +25,8 @@ async fn main() -> Result<()> {
...
-        .and(warp::path::end())
-        .and(bobtimus_filter.clone())
-        .and(warp::body::json())
-        .and_then(routes::create_swap);
-
-    warp::serve(latest_rate.or(create_swap))
-        .run(([127, 0, 0, 1], api_port))
-        .await;
+    let routes = http::create_routes(&bobtimus, bobtimus.rate_service.subscribe());
+    warp::serve(routes).run(([127, 0, 0, 1], api_port)).await;
```
thomaseizinger, line 29: In theory, you could even move the `warp::serve` into `lib.rs` and expose it as `http::start` :)

This is also fine of course! Given that it is a library interface, it feels a bit weird to take all of this "control" away from the binary. However, one could argue that in this case, we don't want the binaries to have control because they should behave identical when it comes to the API.

thomaseizinger (approved):

bobtimus/src/bin/bobtimus.rs
luckysori: Haha nice catch woops

PR #29, opened 2020-12-14
Add coin selection API

luckysori:
A very naive way of using `bdk`'s coin selection algorithm.

There is one thing which is definitely wrong: we're currently setting the max weight of each UTXO to zero because the proper way of calculating it is through a `miniscript` API, namely `Descriptor::max_satisfaction_weight`.

### DIFF
--- a/bobtimus/src/coin_selection.rs
+++ b/bobtimus/src/coin_selection.rs
@@ -1,0 +1,259 @@
+use anyhow::Result;
+use bdk::{
+    bitcoin::Amount,
+    database::{BatchOperations, Database},
+    wallet::coin_selection::CoinSelectionResult,
+    wallet::coin_selection::{BranchAndBoundCoinSelection, CoinSelectionAlgorithm},
+};
+use elements_fun::{ExplicitTxOut, OutPoint};
+
+#[derive(Clone, PartialEq)]
+pub struct Utxo {
+    outpoint: OutPoint,
+    txout: ExplicitTxOut,
+}
+
+impl From<Utxo> for bdk::UTXO {
+    fn from(utxo: Utxo) -> Self {
+        let value = utxo.txout.value.0;
+        let script_pubkey = utxo.txout.script_pubkey.into_bytes();
+        let script_pubkey = bdk::bitcoin::Script::from(script_pubkey);
+
+        Self {
+            outpoint: bdk::bitcoin::OutPoint {
+                txid: bdk::bitcoin::Txid::from_hash(utxo.outpoint.txid.as_hash()),
+                vout: utxo.outpoint.vout,
+            },
+            txout: bdk::bitcoin::TxOut {
+                value,
+                script_pubkey,
+            },
+            script_type: bdk::ScriptType::External,
+        }
+    }
+}
+
+pub fn coin_select(utxos: Vec<Utxo>, target: Amount) -> Result<Vec<Utxo>> {
+    let algorithm = BranchAndBoundCoinSelection::default();
+
+    let bdk_utxos = utxos
+        .iter()
+        .cloned()
+        .map(bdk::UTXO::from)
+        .map(|utxo| (utxo, 0))
+        .collect();
+
+    let CoinSelectionResult {
+        selected: selected_utxos,
+        ..
+    } = algorithm.coin_select(
+        &DummyDb,
+        Vec::new(),
+        bdk_utxos,
+        bdk::FeeRate::default(),
+        target.as_sat(),
+        0.0,
+    )?;
+
+    let selected_utxos = selected_utxos
+        .iter()
+        .map(|bdk_utxo| {
+            utxos
+                .iter()
+                .find(|utxo| {
+                    bdk_utxo.outpoint.txid.as_hash() == utxo.outpoint.txid.as_hash()
+                        && bdk_utxo.outpoint.vout == utxo.outpoint.vout
+                })
+                .expect("same source of utxos")
+        })
+        .cloned()
+        .collect();
+
+    Ok(selected_utxos)
+}
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
+    use elements_fun::{AssetId, ExplicitAsset, ExplicitValue, Script, Txid};
+
+    use super::*;
+
+    #[test]
+    fn trivial_coin_selection() {
+        let utxo = Utxo {
+            outpoint: OutPoint {
+                txid: Txid::default(),
+                vout: 0,
+            },
+            txout: ExplicitTxOut {
+                asset: ExplicitAsset(
+                    AssetId::from_str(
+                        "0000000000000000000000000000000000000000000000000000000000000000",
+                    )
+                    .unwrap(),
+                ),
+                value: ExplicitValue(100_000_000),
+                script_pubkey: Script::new(),
+                nonce: None,
+            },
+        };
+
+        let selection = coin_select(vec![utxo.clone()], Amount::from_sat(90_000_000)).unwrap();
+
+        assert!(selection.len() == 1);
+        assert!(selection.contains(&utxo));
+    }
+}
+
+/// A placeholder for the `database` argument required by
+/// `CoinSelectionAlgorithm::coin_select`, but which is never actually
+/// used in the trait implementation.
+struct DummyDb;
+
+impl Database for DummyDb {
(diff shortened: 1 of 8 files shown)


bobtimus/src/coin_selection.rs
```diff
@@ -0,0 +1,259 @@
...
+    database::{BatchOperations, Database},
+    wallet::coin_selection::CoinSelectionResult,
+    wallet::coin_selection::{BranchAndBoundCoinSelection, CoinSelectionAlgorithm},
+};
+use elements_fun::{ExplicitTxOut, OutPoint};
+
+#[derive(Clone, PartialEq)]
+pub struct Utxo {
+    outpoint: OutPoint,
+    txout: ExplicitTxOut,
```
thomaseizinger, line 13: This will need to be public so we can actually instantiate it outside of this module.

bobtimus/src/coin_selection.rs
```diff
@@ -0,0 +1,259 @@
...
+            txout: bdk::bitcoin::TxOut {
+                value,
+                script_pubkey,
+            },
+            script_type: bdk::ScriptType::External,
+        }
+    }
+}
+
+pub fn coin_select(utxos: Vec<Utxo>, target: Amount) -> Result<Vec<Utxo>> {
```
thomaseizinger, line 36: We will need the fee as an output as well, plus the change amount.

thomaseizinger, line 36: If we accept a `Utxo` here which has an `ExplicitTxOut`, we should probably assert that they are all the same asset. Alternatively, we could inline `value` and `script_pubkey` and document that it needs to be the same asset.

thomaseizinger (changes requested):
Thanks for looking into this @luckysori !

I am not sure I would want to depend on `bobtimus` from the wallet, so this will have to go into a different crate in our workspace.

Other than that, it looks like something I can work with (modulo the inline comments), thank you!

luckysori:
> I am not sure I would want to depend on `bobtimus` from the wallet, so this will have to go into a different crate in our workspace.

Totally, it needs to be somewhere else, but I didn't want to figure out the correct place at this stage 😛

bobtimus/src/coin_selection.rs
luckysori: We can output the recommended fee, but it currently only takes into account the weight of each input and _not_ the weight of whatever will be added to the transaction to satisfy it. Meaning it is an unreliable estimate. Do you think it would be sufficient to assume the highest possible "satisfaction weight"? Or is it worth trying to figure out the value from the script, a bit like miniscript does (maybe there's code for this in bitcoind)?

The change amount should just be the result of adding all the selected UTXOs, minus the estimated fee, but we can return it explicitly if it's preferable.

thomaseizinger: > The change amount should just be the result of adding all the selected UTXOs, minus the estimated fee, but we can return it explicitly if it's preferable.

Fair. I agree that this makes the return values not completely orthogonal but at least in elements, we need to be explicit about the fee which means every caller will have to do this summing.

> We can output the recommended fee, but it currently only takes into account the weight of each input and not the weight of whatever will be added to the transaction to satisfy it. Meaning it is an unreliable estimate. Do you think it would be sufficient to assume the highest possible "satisfaction weight"? Or is it worth trying to figure out the value from the script, a bit like miniscript does (maybe there's code for this in bitcoind)?

At least in the browser-based wallet, we can assume a fixed weight per UTXO because it is always the same script (p2wpkh).

thomaseizinger: We only have standard scripts on both sides, so a fixed value per script pub key should be fine. It is obviously not a general solution (that would require miniscript) but it should be good enough for our usecase.

`Script` has `is_XYZ` classifiers which you should be able to use to decide on a reasonable estimate.

luckysori: Let's hard-code it to the max satisfaction weight of a p2wpkh script then, because we don't need this to be a general purpose coin selection algorithm yet anyway.

> Script has is_XYZ classifiers which you should be able to use to decide on a reasonable estimate.

I've been trying to use them, but some of the variants don't have obvious (to me) max satisfaction weights.

bonomat: > We only have standard scripts on both sides, so a fixed value per script pub key should be fine. It is obviously not a general solution (that would require miniscript) but it should be good enough for our usecase.

The elements-miniscript fork seems to make some progress: https://github.com/sanket1729/elements-miniscript. I don't know if it's usable yet though.

For now we can assume a fixed satisfaction weight because we only support `P2WPK` anyways.

bonomat:
> I am not sure I would want to depend on bobtimus from the wallet, so this will have to go into a different crate in our workspace.

Shall we move it into `elements-fun`?

bobtimus/src/coin_selection.rs
thomaseizinger: > Let's hard-code it to the max satisfaction weight of a p2wpkh script then, because we don't need this to be a general purpose coin selection algorithm yet anyway.
> 
> > Script has is_XYZ classifiers which you should be able to use to decide on a reasonable estimate.
> 
> I've been trying to use them, but some of the variants don't have obvious (to me) max satisfaction weights.

You don't need to handle all of the variants. Witness and non-witness p2pk should be fine. Any other utxos (on bobtimus' side) can be filtered out!

luckysori:
> > I am not sure I would want to depend on bobtimus from the wallet, so this will have to go into a different crate in our workspace.
> 
> Shall we move it into `elements-fun`?

I think I'll just put it where it's gonna be used i.e. frontend wallet. With `elements-fun` hopefully getting merged into `rust-elements` in the future, it should pay off to keep things separate.

bobtimus/src/coin_selection.rs
luckysori: > The elements-miniscript fork seems to make some progress: https://github.com/sanket1729/elements-miniscript. I don't know if it's usable yet though.

That looked promising when I took a look, but the problem that I see is that the UTXOs that we get from esplora or elementsd don't have a descriptor associated with them (AFAIK).

luckysori: > You don't need to handle all of the variants. Witness and non-witness p2pk should be fine. Any other utxos (on bobtimus' side) can be filtered out!

I like that approach: handle the two relevant cases and filter out the rest of the provided UTXOs that don't apply.

thomaseizinger:
> > > I am not sure I would want to depend on bobtimus from the wallet, so this will have to go into a different crate in our workspace.
> > 
> > 
> > Shall we move it into `elements-fun`?
> 
> I think I'll just put it where it's gonna be used i.e. frontend wallet. With `elements-fun` hopefully getting merged into `rust-elements` in the future, it should pay off to keep things separate.

With "go into a different crate" I literally meant, put it into its own crate :)
Isn't bobtimus also gonna need this or how are you selecting from our list of UTXOs?

luckysori:
> With "go into a different crate" I literally meant, put it into its own crate :)

That was the alternative, but I think it's fine to let `elementsd` choose on behalf of Bobtimus (it uses the same algorithm under the hood).

thomaseizinger (approved):
Nice but we need this to compile to wasm :)

waves/wallet/Cargo.toml
```diff
@@ -13,7 +13,7 @@ default = [ "console_error_panic_hook" ]
 [dependencies]
 aes-gcm-siv = { version = "0.9", features = [ "std" ] }
 anyhow = "1"
-bdk = { git = "https://github.com/bitcoindevkit/bdk", default-features = false }
+bdk = { git = "https://github.com/coblox/bdk", default-features = false }
```
thomaseizinger, line 16: Could you not use `[patch.crates-io]` to replace the `miniscript` dependency?

luckysori: Ah, I forget how to use this feature. That might be better!

luckysori: You were right, one fewer fork to maintain.

thomaseizinger: It is already [merged](https://github.com/rust-bitcoin/rust-miniscript/pull/212) into master so you could even depend on their HEAD directly instead of our fork.

luckysori: I tried that, but there's a tricky breaking change in between 4.0 (the version `bdk` depends on) and our commit.

PR #31, opened 2020-12-17
Some wallet refactoring and wallet info component added

bonomat:
A few changes: 

* Refactored wallet update to `updateBalance` and `unlockWallet`. On `unlockWallet` I keep polling the wallet for a new balance
* New UI component: `WalletInfo`
image

@thomaseizinger : we need additional wallet functionality please to withdraw funds. I though of a simple `withdrawall` which withdraws all assets to a given address.


waves/src/App.tsx
```diff
@@ -167,29 +203,49 @@ function App() {
...
-                        btcBalance: 0,
-                        usdtBalance: 0,
                     },
                 });
             } // by default `wallet.exists` is set to false, hence no need to handle
         });
     }, []);
 
+    useEffect(() => {
+        let updateBalanceInterval = setInterval(() => {
```
thomaseizinger: There is a library that provides a hook for this if you don't want to roll your own: https://usehooks-typescript.com/react-hook/use-interval

waves/src/WalletInfo.tsx
```diff
@@ -29,8 +30,11 @@ interface WalletInfoProps {
 export default function WalletInfo({ balance }: WalletInfoProps) {
     const { isOpen, onOpen, onClose } = useDisclosure();
     const btnRef = React.useRef(null);
-    const address = "LTB_addressxtqwseasdas"; // TODO get address from
     const [withDrawAddress, setWithDrawAddress] = React.useState("");
+    const [address, setAddress] = React.useState("LTB_addressxtqwseasdas");
+    getAddress().then((address) => {
+        setAddress(address);
+    });
```
thomaseizinger, line 37: This should be inside a `useEffect` otherwise it will run on every re-render.

thomaseizinger: If you want to make this less stateful, you could also use something like this: https://docs.react-async.com/guide/async-components#more-flexibility-with-useasync

In that case, you would probably want to extract a component for the whole QR code that internally renders something (or nothing) while the `getAddress` is pending.

thomaseizinger: The same pattern can likely be applied in other cases as well!

waves/src/App.tsx
thomaseizinger: I saw in the the next patch that you are actually modifying this further. Might be a good idea to use the hook from the library then.

thomaseizinger (approved):
LGTM modulo the missing `useEffect` / alternative async handling.

waves/src/WalletInfo.tsx
bonomat: Cool stuff, used it in the last patch: [`86d3141` (#31)](https://github.com/comit-network/droplet/pull/31/commits/86d31416072d2ad684dea6035ddfa23d1e80c1b9)

waves/src/App.tsx
bonomat: So much nicer with this hook :)

PR #45, opened 2020-12-18
Introduce "withdraw all" functionality

### LINKED ISSUE #32: Wallet: withdraw all function

thomaseizinger:
Fixes #32.

### DIFF
--- a/waves/wallet/src/wallet.rs
+++ b/waves/wallet/src/wallet.rs
@@ -169,10 +169,10 @@ pub async fn get_address(
 }
 
 pub async fn get_balances(
-    name: String,
+    name: &str,
     current_wallet: &Mutex<Option<Wallet>>,
 ) -> Result<Vec<BalanceEntry>, JsValue> {
-    let wallet = current(&name, current_wallet).await?;
+    let wallet = current(name, current_wallet).await?;
 
     let txouts = get_txouts(&wallet).await?;
 
--- a/waves/wallet/src/wallet.rs
+++ b/waves/wallet/src/wallet.rs
@@ -46,8 +46,6 @@ pub async fn create_new(
         )));
     }
 
-    wallets.add(name.clone());
-
     let params = if cfg!(debug_assertions) {
         // use weak parameters in debug mode, otherwise this is awfully slow
         log::warn!("using extremely weak scrypt parameters for password hashing");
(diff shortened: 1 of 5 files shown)


waves/wallet/src/wallet.rs
```diff
@@ -210,6 +221,258 @@ pub async fn current<'n, 'w>(
...
+        .group_by(|(_, _, unblinded)| unblinded.asset);
+
+    // prepare the data exactly as we need it to create the transaction
+    let txouts_grouped_by_asset = (&txouts_grouped_by_asset)
+        .into_iter()
+        .map(|(asset, txouts)| {
+            let txouts = txouts.collect::<Vec<_>>();
+
+            // calculate the total amount we want to spend for this asset
+            // if this is the native asset, subtract the fee
```
luckysori, line 295: This reminds me that a user can't withdraw without sufficient L-BTC.

thomaseizinger: Yes, for now the UI could just call this function and display the error message? That would say that you need L-BTC.

We can also add an additional check in the UI!

waves/wallet/src/wallet.rs
```diff
@@ -281,6 +544,29 @@ where
         .await
 }
 
+/// Estimate the virtual size of a transaction based on the number of inputs and outputs.
```
luckysori, line 547: :abacus:

waves/wallet/src/wallet.rs
luckysori: I see that the error is handled below.

waves/wallet/src/wallet.rs
```diff
@@ -210,6 +221,258 @@ pub async fn current<'n, 'w>(
...
+                asset,
+                total_input
+            );
+
+            (asset, txins, txout_inputs, to_spend)
+        })
+        .collect::<Vec<_>>();
+
+    // build transaction from grouped txouts
+    let mut transaction = match txouts_grouped_by_asset.as_slice() {
```
luckysori, line 352: I love using subslice patterns!

luckysori (approved):
That `withdraw_everything_to` function looks crazy :eyes:

waves/wallet/src/wallet.rs
thomaseizinger: They are great right? Loving it too!

PR #46, opened 2020-12-18
Add SSE client side

### LINKED ISSUE #39: Frontend: use bobtimus rate service
Frontend should get the rate from bobtimus via SSE

bonomat:
fixes #39


waves/src/App.test.tsx
```diff
@@ -1,9 +1,26 @@
 import { render, screen } from "@testing-library/react";
 import React from "react";
+import { Listener, Source, SSEProvider } from "react-hooks-sse";
 import App, { AssetType, reducer } from "./App";
 
-test("renders create new wallet", () => {
-    render(<App />);
+// implementation of customSource does not matter but functions need to be there
```
thomaseizinger, line 6: `DummySource` might be a good name then!

waves/src/App.tsx
```diff
@@ -229,6 +223,18 @@ function App() {
         state.wallet.status.loaded ? 5000 : null,
     );
 
+    state.rate = useSSE("rate", {
```
thomaseizinger, line 226: Uhm, why are you overwriting the `state` variable here? With reducers, the `state` should be read-only!

waves/src/App.tsx
```diff
@@ -229,6 +223,18 @@ function App() {
...
+        ask: 19133.74,
+        bid: 19133.74,
+    });
+
+    useEffect(() => {
+        dispatch({
+            type: "UpdateBetaAmount",
+            value: state.rate,
+        });
+    }, [state.rate]);
```
thomaseizinger, line 236: Non-urgent feedback, more of a long-term thing to consider:

It is a bit odd that you have to trigger another effect here based on the rate change because we already have a reducer-like construct within `useSSE`.

I think this stems from your reducer calculating both amounts. That is in the end more state than you should keep in your reducer (i.e. it is not normalized because one field depends on the other).

I feel it should be possible to model this in a more functional way where the only thing you store are the input values (and probably which field the user touched last) and you dynamically compute the other inputs field based on the rate. By dynamically computing I mean outside of the reducer, simply as part of the render function.

That would allow you to remove the rate from the reducer and you can just use the variable returned from `useSSE` which in turn makes this `useEffect` go away!

A component (in particular the render function but you only have function components so here it is equivalent) is always a pure function of data (props + state) to a view. Any kind of side-effect is either outside or modeled through hooks.
It makes things a lot easier to understand if you use side-effects only for the things that have to be side-effects. For example, from a domain-perspective, one of the input's values is always a pure function of the rate + the other input. As such, computing its value could be done as part of the render function.

waves/src/index.tsx
```diff
@@ -1,18 +1,18 @@
...
-import { Provider as RateServiceProvider } from "./hooks/RateService";
 import "./index.css";
 import reportWebVitals from "./reportWebVitals";
 import theme from "./theme";
 
 ReactDOM.render(
     <React.StrictMode>
         <ChakraProvider theme={theme}>
-            <RateServiceProvider value="http://getbestrate.com">
+            <SSEProvider endpoint="http://localhost:3031/rate/lbtc-lusdt">
```
thomaseizinger, line 13: This should be a relative URL otherwise it will not work in production.

thomaseizinger (changes requested):
Couple of minor comments, otherwise LGTM

waves/src/index.tsx
bonomat: True, this needs some more investigation. Unfortunately it does not work with relative urls. My assumption is that it's because of the proxy. The server receives the request and keeps sending, but the client does not get the response.

bonomat: ah, it could also because of stuff running in the docker container ... not sure ...

waves/src/App.tsx
bonomat: I'm not sure I follow. 

The beta amount is always `alpha`*`rate` and cannot be edited. This was changed when you mentioned it the first time :) 

So to the rate, `useSSE` returns the rate. How should I trigger an update of the `beta`-amount if not with the reducer?

waves/src/App.tsx
bonomat: see discussion below.

waves/src/App.tsx
bonomat: Ah, I think I have an idea...

bonomat: checkout latest commit: [`9d36876` (#46)](https://github.com/comit-network/droplet/pull/46/commits/9d368769592eeb7bc9403e2c466c3f667fbd9091). I'm not sure how to get rid of `state.rate = useSSE` as I store the rate in the state.

thomaseizinger: The last patch looks almost like what I meant :)

Why does the rate have to be in the state though?

Just make it a local variable of your component!

thomaseizinger: If the rate has to be in the state because you need it in the reducer for handling other actions, you should introduce the action again that dispatches the rate.

waves/src/App.tsx
```diff
@@ -243,19 +249,27 @@ function App() {
...
-                                    <ExchangeIcon dispatch={dispatch} />
+                                    <ExchangeIcon
+                                        onClick={() =>
+                                            dispatch({
+                                                type: "SwapAssetTypes",
+                                                value: {
+                                                    betaAmount,
+                                                },
+                                            })}
+                                    />
```
bonomat, line 260: Personally I prefer extracting functions for this which makes the html code more readable.

thomaseizinger: I am not sure I fully agree. I find React code easier to read if components are extracted in a particular way. It is quite nice to know that clicking this button will swap the assets _without_ having to go into the component and check what it does.

This way, all the state and how / when it is modified is within a single file. If you start extracting functions for this, you have to jump around again between fn definition vs UI definition.

thomaseizinger: One way to make this more concise is to extract helper functions for creating actions as in:

```ts
function swapAssetTypes(betaAmount: number) {
	// ...
}
```

so you can do:

```ts
dispatch(swapAssetTypes(betaAmount))
```

However, that is also quite the boilerplate esp. if things are still changing a lot so :man_shrugging:

thomaseizinger: I'd probably first go ahead and remove the `.value` from all actions and just embed whatever type you want directly. That would also remove 2 lines of code from creating this action.

waves/src/App.test.tsx
```diff
@@ -150,25 +150,34 @@ test("Swap asset types", () => {
...
         reducer(initialState, {
             type: "SwapAssetTypes",
+            value: {
+                betaAmount: initialState.alpha.amount,
+            },
         }).beta,
     ).toBe(initialState.alpha.type);
 
     // amounts should be flipped as well
+    let betaAmount = calculateBetaAmount(initialState.alpha.type, initialState.alpha.amount, initialState.rate);
```
thomaseizinger, line 172: This still accesses the `rate` from the state but in the actual code, the state no longer has the rate.

waves/src/App.test.tsx
```diff
@@ -150,25 +150,34 @@ test("Swap asset types", () => {
         beta: AssetType.USDT,
     };
 
-    let newValue = AssetType.BTC;
     expect(
         reducer(initialState, {
             type: "SwapAssetTypes",
+            value: {
+                betaAmount: initialState.alpha.amount,
```
thomaseizinger, line 157: Not sure if this test makes a whole lot of sense if you pass in the original value of `alpha.amount`?

The reducer no longer calls `calculateBetaAmount`. Hence, the current behaviour can be tested with a much simpler and easier to understand test (simply assert that the value passed in the action gets set to `alpha.amount`).

bonomat: For me this makes a whole lot of sense as it shows that alpha and beta amounts are swapped afterwards. :) 

Let me know what you don not understand here please and we can change the test. :)

thomaseizinger: > For me this makes a whole lot of sense as it shows that alpha and beta amounts are swapped afterwards. :)

How much value does the assertion have if there is not observable difference because the numbers are the same? :D

The reducer sets `alpha.amount` to `betaAmount`. If you pass `state.alpha.amount` as `betaAmount` there is no effective change of the value of `alpha.amount`.

Hence the assertion is weak because removing the assignment from within the reducer would not actually break the test.

`SwapAssetTypes` does two concrete things:

- swap the types of the amounts
- assign `betaAmount` to `state.alpha.amount`

Given that, a clear test should:

1. Initialize the state (ideally in a concise way) that shows which elements of the state we care about: `alpha.type`, `alpha.amount` & `beta.type`
2. Dispatch `SwapAssetTypes` with a hard-coded number where it is clearly visible that it is different from `alpha.amount`
3. Assert that the asset types are flipped and that `betaAmount` is the new `alpha.amount`

bonomat: > Given that, a clear test should:

> 
>     Initialize the state (ideally in a concise way) that shows which elements of the state we care about: alpha.type, alpha.amount & beta.type
>     Dispatch SwapAssetTypes with a hard-coded number where it is clearly visible that it is different from alpha.amount
>     Assert that the asset types are flipped and that betaAmount is the new alpha.amount


Imho that's precisely what I'm doing. :) 

1) First few lines of the test setup the sate: 
https://github.com/comit-network/droplet/blob/d7c99d65dd47e8f1fda67df9422aa090d5f2bee7/waves/src/App.test.tsx#L136-L143

2) Dispatched the `SwapAssetType` with a fixed amount (I slightly changed it in the last patch)

https://github.com/comit-network/droplet/blob/d7c99d65dd47e8f1fda67df9422aa090d5f2bee7/waves/src/App.test.tsx#L151

3) Assert that the types are flipped and the `betaAmount` as specified in the event is assigned to `alpha.amount`.

bonomat:
@thomaseizinger : any more feedback or can we go ahead and merge this?

thomaseizinger (approved):
Yeah looking good!

PR #47, opened 2020-12-18
Add script to spin up local development environment

### LINKED ISSUE #34: Infrastructure: Development environment with local esplora and elementd
Esplora URL for the wallet can be set using `ESPLORA_URL` env variable at **build** time: https://github.com/comit-network/droplet/blob/96687ef6f0c9acd3795cfcab03bb6fc427c5ddab/waves/wallet/src/esplora.rs#L9

luckysori:
Fixes #34.

Requires having [nigiri](https://github.com/vulpemventures/nigiri) in the path.

Also adds a button to frontend wallet which funds it with some L-BTC when clicked.

Missing things:

- Not spinning up all the unnecessary containers related to bitcoind.
- Tear down: 
  At the moment I'm just running the following commands when I want to reset the environment:
  ```
  nigiri --datadir $(realpath .nigiri) stop --delete
  pkill bobtimus
  ```


waves/src/WalletInfo.tsx
```diff
@@ -133,15 +140,20 @@ export default function WalletInfo({ balance }: WalletInfoProps) {
...
                                     </HStack>
                                 </VStack>
+                                <VStack bg="gray.100" align="center" borderRadius={"md"} p={1}>
+                                    <Button
+                                        size="md"
+                                        variant="wallet_button"
+                                        onClick={fundWallet}
+                                    >
+                                        Fund wallet
+                                    </Button>
```
thomaseizinger, line 150: Can we have this button in development mode only please? See here for how to do this: https://create-react-app.dev/docs/adding-custom-environment-variables/

create-liquid-app
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
thomaseizinger, line 1: I think what is script does is more align with `react-scripts` rather than `create-react-app`. As such, you might want to call this maybe `liquid-scripts` or more specifically `waves-scripts`  (given that it deals with bobtimus)? :)

create-liquid-app
```diff
@@ -0,0 +1,36 @@
...
+set -e
+
+# ensure that .log directory exists
+readonly log_dir="./.log"
+mkdir -p $log_dir
+
+readonly nigiri_config_path=$(realpath ./.nigiri)
```
thomaseizinger, line 8: Are you planning to add this directory to the repository? Because if not then I don't see a reason why we couldn't stick with the default.

waves/package.json
```diff
@@ -35,7 +35,7 @@
     "web-vitals": "^0.2.4"
   },
   "scripts": {
-    "start": "react-app-rewired start",
+    "start": "PORT=3004 react-app-rewired start",
```
thomaseizinger, line 38: Why can this not be done in the script instead?

thomaseizinger: Also, would `REACT_APP_PORT` work here to be less ambiguous?

create-liquid-app
```diff
@@ -0,0 +1,36 @@
...
+RUST_LOG=info cargo run --bin fake_bobtimus -- \
+        --elementsd http://$elementsd_rpc_user:$elementsd_rpc_password@127.0.0.1:$LIQUID_NODE_PORT \
+        --usdt $mint_asset_id > $log_dir/bobtimus 2>&1 &
+bobtimus_pid=$!
+
+echo "Starting bobtimus with pid $bobtimus_pid"
+
+# start frontend
+readonly chopsticks_url="http://127.0.0.1:$LIQUID_CHOPSTICKS_PORT"
+cd waves && ESPLORA_URL=$chopsticks_url REACT_APP_ESPLORA_URL=$chopsticks_url yarn start
```
thomaseizinger, line 36: I am not sure how often we will need to start / stop things but if that turns out to be the case (with your emacs setup for example) then it might be better if this script where to export _functions_ to the current shell instead of running the actual commands.

I am thinking of something like:

- `start_bobtimus`
- `stop_bobtimus`
- `start_frontend`
- etc

That way, running the script will just the define the functions correctly and it is quite easy to start & stop things :)

thomaseizinger (approved):
Nice work! Left some comments with some ideas :)

create-liquid-app
luckysori: Hehe, thanks for the suggestion. The name was in jest, but that makes sense!

create-liquid-app
luckysori: You're right. Originally I anticipated having to modify some config files, but I ended up not needing to, so we can simplify this.

create-liquid-app
```diff
@@ -0,0 +1,36 @@
...
+
+echo "Minted $mint_amount USDt with asset ID $mint_asset_id to elementsd wallet address $mint_address"
+
+# start bobtimus
+readonly elementsd_rpc_user="admin1"
+readonly elementsd_rpc_password="123"
+RUST_LOG=info cargo run --bin fake_bobtimus -- \
+        --elementsd http://$elementsd_rpc_user:$elementsd_rpc_password@127.0.0.1:$LIQUID_NODE_PORT \
+        --usdt $mint_asset_id > $log_dir/bobtimus 2>&1 &
+bobtimus_pid=$!
```
bonomat, line 30: I'm afraid that personally I'll need to run this inside a docker container to be able to have waves communicate with bobtimus.

waves/package.json
luckysori: > Why can this not be done in the script instead?

I tried different combinations, but only this worked. I couldn't figure out how to forward this env variable, but it was getting late.

create-liquid-app
luckysori: > with your emacs setup for example

I managed to fix that!

I like your idea, although I think there's still value in having one command that starts everything up, since the order in which things are started matters. That one could be a composition of the functions that you suggest, which we can still expose separately.

waves/src/WalletInfo.tsx
```diff
@@ -39,12 +39,24 @@ export default function WalletInfo({ balance }: WalletInfoProps) {
         console.log("Withdrawing to " + withDrawAddress);
     };
 
+    async function fundWallet(): Promise<any> {
```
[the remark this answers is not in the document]
luckysori: As far as I can tell `nigiri` does not support "fauceting" other assets. One can mint to an address, but we already mint L-USDt for a Bobtimus address, so minting again will produce a different asset.

create-liquid-app
luckysori: Is there anything that we can do so that you don't need to run this in a docker container?

waves/src/WalletInfo.tsx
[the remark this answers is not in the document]
thomaseizinger: FakeBobtimus could expose an endpoint that funds your wallet.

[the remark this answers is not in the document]
thomaseizinger: We could use that for both assets actually.

create-liquid-app
bonomat: Maybe nothing :) , I think I tried to connect to localhost from within a container before and it worked well: https://stackoverflow.com/questions/24319662/from-inside-of-a-docker-container-how-do-i-connect-to-the-localhost-of-the-mach

waves-scripts
```diff
@@ -0,0 +1,111 @@
...
+            --elementsd http://$elementsd_rpc_user:$elementsd_rpc_password@127.0.0.1:$LIQUID_NODE_PORT \
+            --usdt $1 > $log_dir/bobtimus 2>&1 &
+    bobtimus_pid=$!
+
+    # TODO: Make sure that bobtimus was actually started up successfully before recording its pid
+
+    # save bobtimus pid to file for future teardown
+    echo $bobtimus_pid > .bobtimus
+
+    echo "Starting bobtimus with pid $bobtimus_pid"
```
thomaseizinger, line 69:
```suggestion
    echo "Started bobtimus with pid $bobtimus_pid"
```

waves-scripts
```diff
@@ -0,0 +1,111 @@
...
+    __read_env_file
+
+    local -r elementsd_rpc_user="admin1"
+    local -r elementsd_rpc_password="123"
+    RUST_LOG=info cargo run --bin fake_bobtimus -- \
+            --elementsd http://$elementsd_rpc_user:$elementsd_rpc_password@127.0.0.1:$LIQUID_NODE_PORT \
+            --usdt $1 > $log_dir/bobtimus 2>&1 &
+    bobtimus_pid=$!
+
+    # TODO: Make sure that bobtimus was actually started up successfully before recording its pid
```
thomaseizinger, line 64: curl has a flag to make sure the request is a 2xx, you can use that to ping bobtimus on the rate endpoint and check if it is up!

thomaseizinger, line 64: curl has a flag to make sure the request is a 2xx, you can use that to ping bobtimus on the rate endpoint and check if it is up!

thomaseizinger (approved):

waves-scripts
luckysori: That's a good idea! Then I can really be sure that Bob has started and can apply your suggestion above.

waves/package.json
luckysori: > Also, would `REACT_APP_PORT` work here to be less ambiguous?

The environment variable that is expected here is `PORT`, `REACT_APP_PORT` doesn't do anything :(

waves/src/WalletInfo.tsx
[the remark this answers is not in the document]
luckysori: Let's do this later, but I agree!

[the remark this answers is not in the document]
tiero: Found this accidentally, I don't know how :)  We had this on the pipeline from a bit, but now you can. Pull nigiri `0.0.10`


```sh
curl -X POST --data '{"address": "LiquidAddr", "asset": "LiquidAsset", "amount": 0.02}' http://localhost:3001/faucet
```

or via CLI

```
nigiri faucet --liquid <address> [amount] [asset]
```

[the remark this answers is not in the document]
bonomat: Thanks @tiero

PR #53, opened 2020-12-21
Upload hfuzz_workspace in case the fuzzer fails

### LINKED ISSUE #52: Extend CI to upload fuzzer results in case it fails

thomaseizinger:
In order to debug the fuzzer locally, once has to load the crash
file in the debugger. As such, we upload the whole fuzzer workspace
as an artefact so it can be downloaded from the CI job.

Fixes #52.


thomaseizinger:
Getting this one in now in case the fuzzer fails again on another PR.

PR #56, opened 2020-12-21
Add fund L-USDt button to frontend on dev mode

luckysori:


waves/src/WalletInfo.tsx
```diff
@@ -143,9 +155,19 @@ export default function WalletInfo({ balance }: WalletInfoProps) {
                                         <Button
                                             size="md"
                                             variant="wallet_button"
-                                            onClick={fundWallet}
+                                            onClick={fundLbtc}
+                                        >
+                                            Fund L-BTC
```
thomaseizinger, line 160: Can we guard this button as well?

waves/src/WalletInfo.tsx
```diff
@@ -39,7 +39,7 @@ export default function WalletInfo({ balance }: WalletInfoProps) {
         console.log("Withdrawing to " + withDrawAddress);
     };
 
-    async function fundWallet(): Promise<any> {
+    async function fundLbtc(): Promise<any> {
```
thomaseizinger, line 42: I think we should just unify these.

Bobtimus can easily fund both of them at the same time :)
Maybe even with the same button / endpoint?

thomaseizinger (approved):

thomaseizinger (approved):

PR #59, opened 2020-12-22
Use action to install docker for mac

bonomat:


bonomat:
Just realized that the PR ended up here. I only wanted to play around in my own fork. Closing this for now.

PR #66, opened 2020-12-22
Call withdraw_all from WalletInfo

### LINKED ISSUE #69: Wallet: `WithdrawAll` Fails
Broadcasting the transaction failed with: 

```
sendrawtransaction RPC error: {"code":-26,"message":"bad-txns-in-ne-out, value in != value out (code 16)"}
```

bonomat:
This is not working yet unfortunately. 
Fixes #69


waves/src/wasmProxy.ts
```diff
@@ -40,3 +40,8 @@ export async function getBalances(): Promise<BalanceEntry[]> {
     const { get_balances } = await import("./wallet");
     return get_balances(WALLET_NAME);
 }
+
+export async function withDrawAll(address: string): Promise<String> {
```
thomaseizinger, line 44: I would prefer to keep the names consistent everywhere but at the very least, it should be `withdraw` :)

thomaseizinger (commented):
Do you mind putting the commit where you hook up everything as the last one of this PR? Currently it is the 2nd one so we have a somewhat "broken" history.

bonomat (commented):
Can't approve it as I'm the author. But LGTM!

waves/src/WalletInfo.tsx
```diff
@@ -31,12 +31,13 @@ interface WalletInfoProps {
...
+    const [withdrawAddress, setWithdrawAddress] = React.useState("");
 
-    const handleWithdrawAddress = (event: ChangeEvent<HTMLInputElement>) => setWithDrawAddress(event.target.value);
+    const handleWithdrawAddress = (event: ChangeEvent<HTMLInputElement>) => setWithdrawAddress(event.target.value);
 
-    const withdraw = () => {
-        console.log("Withdrawing to " + withDrawAddress);
+    const withdraw = async () => {
+        let txId = await withdrawAll(withdrawAddress);
+        console.log("Withdrew everything. Resulting txId: {}", txId);
```
thomaseizinger, line 40: I guess this can go into a follow-up but I think we should display a little popup and point the user to the transaction on the blockchain explorer. Should probably also be configurable at build-time which explorer it uses.

thomaseizinger (approved):

PR #74, opened 2020-12-23
Add `sign_and_send_swap_transaction` API to wallet 

thomaseizinger:


thomaseizinger:
The fuzzer failure is dealt with in https://github.com/comit-network/droplet/pull/86.

elements-fun/src/transaction.rs
```diff
@@ -327,6 +327,30 @@ pub struct PegoutData<'txo> {
     pub extra_data: Vec<&'txo [u8]>,
 }
 
+/// These constants have been reverse engineered through the following transactions:
+///
+/// https://blockstream.info/liquid/tx/a17f4063b3a5fdf46a7012c82390a337e9a0f921933dccfb8a40241b828702f2
+/// https://blockstream.info/liquid/tx/d12ff4e851816908810c7abc839dd5da2c54ad24b4b52800187bee47df96dd5c
+/// https://blockstream.info/liquid/tx/47e60a3bc5beed45a2cf9fb7a8d8969bab4121df98b0034fb0d44f6ed2d60c7d
```
bonomat, line 334: Did you see that these 3 are `p2sh` transactions? 

An improvement could be to do 3 transactions with the mentioned combinations and record the vbytes.

bonomat (approved):
Beautiful commit messages.

elements-fun/src/transaction.rs
thomaseizinger: Note: This code was only moved in this PR, not added.

The majority of the size comes from the range proof which is almost a kilobyte. I'd expect the size of the script to therefore play a negligible role.

bobtimus/src/lib.rs
```diff
@@ -154,40 +148,52 @@ impl<R, RS> Bobtimus<R, RS> {
...
+        )?;
 
-        let message = bob1.compose(signer).await?;
+        let transaction = swap::bob_create_transaction(
+            &mut self.rng,
+            &self.secp,
+            alice,
+            bob,
+            self.btc_asset_id,
+            Amount::from_sat(1), // TODO: Make this dynamic once there is something going on on Liquid
```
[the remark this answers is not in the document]
thomaseizinger: I know about the minimum but unfortunately we can't express it with `Amount`.

bobtimus/src/lib.rs
```diff
@@ -154,40 +148,52 @@ impl<R, RS> Bobtimus<R, RS> {
                             })?
                             .clone();
 
-                        Result::<_, anyhow::Error>::Ok((txin, txin_as_txout, blinding_key))
+                        Result::<_, anyhow::Error>::Ok(swap::Input {
+                            tx_in,
+                            tx_out,
```
luckysori, line 153: :upside_down_face: Super nitpicky, but we almost exclusively use `txin` and `txout` in this workspace.

swap/src/lib.rs
```diff
@@ -1,462 +1,307 @@
...
-            let sig = secp.sign(&Message::from(sighash), &input_sk);
-
-            let mut serialized_signature = sig.serialize_der().to_vec();
-            serialized_signature.push(SigHashType::All as u8);
-
-            vec![serialized_signature, input_pk_bob.serialize().to_vec()]
-        };
+    pub fn calculate_change_amount(
+        &self,
+        other_receive_asset: AssetId,
```
luckysori, line 268: :question: Maybe `send_asset`?

luckysori (approved):
Really nice effort. I wholeheartedly agree with the TODO indicating that we should replace the `swap` code with PSET asap.

swap/src/lib.rs
thomaseizinger: I am indifferent and lazy so I will just leave it at that :D

PR #75, opened 2020-12-23
Embed waves in bobtimus

### LINKED ISSUE #33: Bobtimus: Embed frontend and serve via warp

bonomat:
I'm not sure if this is the best way to do this but it works. 

Bobtimus embeds the content of wave's build folder during build time. Which means, if the folder is not present, the build fails. I opted for checking this folder in and ignoring its content to enforce that this folder exists. 

Resolves #33

### DIFF
--- a/bobtimus/src/http.rs
+++ b/bobtimus/src/http.rs
@@ -1,9 +1,17 @@
 use elements_fun::secp256k1::rand::{CryptoRng, RngCore};
 use futures::{Stream, StreamExt};
+use rust_embed::RustEmbed;
 use std::convert::Infallible;
-use warp::{filters::BoxedFilter, Filter, Rejection, Reply};
+use warp::{
+    filters::BoxedFilter, http::header::HeaderValue, path::Tail, reply::Response, Filter,
+    Rejection, Reply,
+};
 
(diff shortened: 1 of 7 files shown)


waves/build/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
bonomat, line 4:
```suggestion
!.gitignore

```

bobtimus/src/http.rs
```diff
@@ -13,7 +21,8 @@ where
     R: RngCore + CryptoRng + Clone + Send + Sync + 'static,
     RS: LatestRate + Clone + Send + Sync + 'static,
 {
-    let health = warp::get().and(warp::path::end()).map(warp::reply);
```
thomaseizinger, line 16: Are we dumping the content of index.html in the terminal now in waves script? Would be nice if not :)

thomaseizinger (approved):

bobtimus/src/http.rs
bonomat: Fixed in the last commit ;)

PR #85, opened 2021-01-05
Update bobtimus api and extract helper in waves

### LINKED ISSUE #77: Add api prefix to bobtimus endpoints

bonomat:
This PR adds `api` prefix to all Bobtimus API. I've extracted a helper file `Bobtimus.tsx` in waves which should serve as a onestop place to interact with Bobtimus.

It also enables all cors headers for `fake_bobtimus` which simplifies development. 

Resolves #77

### DIFF
--- a/bobtimus/src/bin/fake_bobtimus.rs
+++ b/bobtimus/src/bin/fake_bobtimus.rs
@@ -43,9 +43,7 @@ async fn main() -> Result<()> {
     let cors = warp::cors().allow_any_origin();
 
     let faucet = warp::post()
-        .and(warp::path!("api" / "faucet" / ..))
-        .and(warp::path::param())
-        .and(warp::path::end())
+        .and(warp::path!("api" / "faucet" / Address))
         .and(bobtimus_filter)
         .and_then(faucet);
 
--- a/bobtimus/src/bin/fake_bobtimus.rs
+++ b/bobtimus/src/bin/fake_bobtimus.rs
@@ -39,6 +39,9 @@ async fn main() -> Result<()> {
         let bobtimus = bobtimus.clone();
         move || bobtimus.clone()
     });
+
+    let cors = warp::cors().allow_any_origin();
+
     let faucet = warp::post()
         .and(warp::path("faucet"))
         .and(warp::path::param())
@@ -46,7 +49,7 @@ async fn main() -> Result<()> {
         .and(bobtimus_filter)
         .and_then(faucet);
 
-    warp::serve(routes.or(faucet))
+    warp::serve(routes.or(faucet).with(cors))
         .run(([127, 0, 0, 1], api_port))
         .await;
 
--- a/bobtimus/src/bin/fake_bobtimus.rs
+++ b/bobtimus/src/bin/fake_bobtimus.rs
@@ -43,7 +43,7 @@ async fn main() -> Result<()> {
     let cors = warp::cors().allow_any_origin();
 
     let faucet = warp::post()
-        .and(warp::path("faucet"))
+        .and(warp::path!("api" / "faucet" / ..))
         .and(warp::path::param())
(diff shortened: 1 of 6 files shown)


waves/src/Bobtimus.tsx
```diff
@@ -0,0 +1,34 @@
...
+export async function postSellPayload(payload: CreateSwapPayload) {
+    let res = await fetch("/api/swap/lbtc-lusdt/sell", {
+        method: "POST",
+        headers: {
+            "Content-Type": "application/json",
+            Accept: "application/json",
+        },
+        body: JSON.stringify(payload),
+    });
+    console.log(res);
```
thomaseizinger, line 20: This looks like a left-over!

waves/src/Bobtimus.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { CreateSwapPayload } from "./wasmProxy";
+
+export async function fundAddress(address: string): Promise<any> {
+    await fetch("/api/faucet/" + address, {
+        method: "POST",
+    });
+}
+
+export async function postSellPayload(payload: CreateSwapPayload) {
+    let res = await fetch("/api/swap/lbtc-lusdt/sell", {
```
thomaseizinger, line 12: This route doesn't match with what bobtimus listens on? It also didn't match before but I just saw it now!

waves/src/index.tsx
```diff
@@ -1,18 +1,16 @@
...
 import reportWebVitals from "./reportWebVitals";
 import theme from "./theme";
 
 ReactDOM.render(
     <React.StrictMode>
         <ChakraProvider theme={theme}>
-            <SSEProvider endpoint="/rate/lbtc-lusdt">
-                <App />
-            </SSEProvider>
+            <BobtimusRateProvider children={<App />} />
```
thomaseizinger, line 13: `children` are usually not passed as literal props. You can do this instead:

```tsx
<BobtimusRateProvider>
	<App />
</BobtimusRateProvider>
```

bobtimus/src/bin/fake_bobtimus.rs
```diff
@@ -43,7 +43,7 @@ async fn main() -> Result<()> {
     let cors = warp::cors().allow_any_origin();
 
     let faucet = warp::post()
-        .and(warp::path("faucet"))
+        .and(warp::path!("api" / "faucet" / ..))
```
thomaseizinger, line 46: I find it confusing to mix the macro-based filter (`path!`) with normal ones and it seems to require you to do this `..` stuff.

If you want to use the macro, you [should](https://docs.rs/warp/0.2.5/warp/macro.path.html#example) be able to do this:

```rust
warp::path!("api" / "faucet" / Address)
```

Alternatively, you should be able to use two, chained `path()` functions.

thomaseizinger (approved):

waves/src/Bobtimus.tsx
bonomat: Maybe I'm missing something, but I think it is correct: 
https://github.com/comit-network/droplet/blob/3c1b8e3cd804d45981b7fecd9d41615e516a1235/bobtimus/src/http.rs#L35-L40

bobtimus/src/bin/fake_bobtimus.rs
bonomat: cool, works.

waves/src/index.tsx
bonomat: Ah, I think now I got it working .

waves/src/Bobtimus.tsx
thomaseizinger: Ah my bad, I was looking at the `rate` endpoint :)

thomaseizinger (approved):

PR #87, opened 2021-01-05
Decompose Bob's transaction into our unblinded inputs and outputs

luckysori:
I decided to change the serialisation format for the transaction because deserialising the JSON was failing on waves wallet. I could not reproduce the error via unit tests ("malformed pedersen commitment") so I went for a quick fix. It may be worth it to get to bottom of it.

### DIFF
--- a/bobtimus/src/http.rs
+++ b/bobtimus/src/http.rs
@@ -66,7 +66,7 @@ where
         .await
         .map(|transaction| serialize_hex(&transaction))
         .map_err(|e| {
-            log::error!("{}", e);
+            log::error!("Failed to handle create swap: {}", e);
             warp::reject::reject()
         })
 }
--- a/waves/wallet/src/wallet/decompose_transaction.rs
+++ b/waves/wallet/src/wallet/decompose_transaction.rs
@@ -1,0 +1,74 @@
+use crate::wallet::{current, get_txouts, Wallet};
+use anyhow::{Context, Result};
+use elements_fun::{secp256k1::SECP256K1, AssetId, Transaction, TxOut, UnblindError};
+use futures::lock::Mutex;
+use serde::Serialize;
+use wasm_bindgen::JsValue;
+
+#[derive(Serialize)]
+pub struct TransactionElements {
+    our_inputs: Vec<(AssetId, u64)>,
+    our_outputs: Vec<(AssetId, u64)>,
+}
+
+pub async fn decompose_transaction(
+    name: String,
+    current_wallet: &Mutex<Option<Wallet>>,
+    transaction: Transaction,
+) -> Result<TransactionElements, JsValue> {
+    let wallet = current(&name, current_wallet).await?;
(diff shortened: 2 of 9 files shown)


bobtimus/src/http.rs
```diff
@@ -59,14 +59,16 @@ where
         warp::reject::reject()
     })?;
 
-    bobtimus
-        .handle_create_swap(payload)
-        .await
-        .map(|message1| warp::reply::json(&message1.transaction))
```
thomaseizinger, line 65: Yeah I don't know why we ever serialized this a JSON here? I've changed this to the hex-representation here: https://github.com/comit-network/droplet/pull/74

waves/wallet/src/wallet/decompose_transaction.rs
```diff
@@ -0,0 +1,77 @@
...
+    transaction: Transaction,
+) -> Result<TransactionElements, JsValue> {
+    let wallet = current(&name, current_wallet).await?;
+    let blinding_key = wallet.blinding_key();
+
+    let our_inputs = get_txouts(&wallet, |utxo, ref txout| {
+        transaction
+            .input
+            .iter()
+            .find_map(|txin| {
```
thomaseizinger, line 29: I assume you want to `filter_map` here?

thomaseizinger: I don't understand how this is actually giving you a `Vec` :man_shrugging:

thomaseizinger (changes requested):
Nice work!

Two minor nits :)

waves/wallet/src/wallet/decompose_transaction.rs
luckysori: You wrote `get_txouts`!! :stuck_out_tongue: The way I understand it is that every wallet txout will be filtered by checking if it is part of Bob's transaction, which is why if we find it in Bob's transaction we "add" it to the inputs that will be returned by `get_txouts`.

bobtimus/src/http.rs
luckysori: I saw your change! Is bincode okay, or should we stick to hex? I first considered hex but I gave up after trying to deserialise it into a `Transaction` and failing.

waves/wallet/src/wallet/decompose_transaction.rs
thomaseizinger: I am confused. From my understanding, `find_map` will only give you the first one and not a list which is why I don't understand how `our_inputs` is a `Vec` again in the end. Is there some magic going on in `collect`?

I think you want to be using `filter_map`, otherwise you will always get a max one input.

bobtimus/src/http.rs
bonomat: I would prefer hex as it seems the _natural_ way to serialize a transaction. Happy to have a look at the deserializing problem together.

thomaseizinger (approved):

PR #98, opened 2021-01-06
Return detailed error message to client using Http Error Codes.

### LINKED ISSUE #92: Bobtimus erronously returns a 404 if creating a swap fails
Related to https://github.com/comit-network/droplet/issues/91.

Getting a 404, should probably be a 400 or 500, depending on what the root-cause is.

bonomat:
I followed the approach we took in comit-rs (and c&p problem.rs).

The error looks like this: 

Server:
```bash
[2021-01-06T03:32:52Z INFO  bobtimus::problem] route failed because failed to calculate change amount for alice: The inputs contain an AssetId != 5ac9f65c0efcc4775e0baec4ec03abdde22473cd3cf33c0419ca290e0751b225.
```

Client: 
image

Resolves #92

### DIFF
--- a/swap/src/lib.rs
+++ b/swap/src/lib.rs
@@ -239,6 +239,18 @@ pub struct Actor {
     receive_amount: Amount,
 }
 
+#[derive(thiserror::Error, Debug, Clone, Copy)]
+#[error("The inputs contain an AssetId != {0}.")]
+pub struct InvalidAssetTypes(pub AssetId);
+
+#[derive(thiserror::Error, Debug, Clone, Copy)]
+#[error("Amount_in ({0}) < amount_out ({1})")]
+pub struct InputAmountTooSmall(pub u64, pub u64);
+
+#[derive(thiserror::Error, Debug, Clone, Copy)]
+#[error("Change_amount ({0}) < fee ({1})")]
+pub struct ChangeAmountTooSmall(pub u64, pub u64);
+
 impl Actor {
     pub fn new<C>(
         secp: &Secp256k1<C>,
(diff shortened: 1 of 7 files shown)


swap/src/lib.rs
```diff
@@ -288,16 +300,20 @@ impl Actor {
...
+            change_amount
+                .checked_sub(fee_amount)
+                .with_context(|| {
+                    format!(
+                        "change_amount ({}) < fee ({})",
+                        change_amount.as_sat(),
+                        fee_amount.as_sat()
+                    )
+                })
+                .map_err(|_| ChangeAmountTooSmall(change_amount.as_sat(), fee_amount.as_sat()))?
```
thomaseizinger, line 316: Same here.

swap/src/lib.rs
```diff
@@ -288,16 +300,20 @@ impl Actor {
                     amount_in,
                     other_receive_amount.as_sat()
                 )
-            })?;
+            })
+            .map_err(|_| InputAmountTooSmall(amount_in, other_receive_amount.as_sat()))?;
```
thomaseizinger, line 304: 2 things:

1. Unless there is a specific reason not to (like the source error not implementing `StdError`), you should always use `context` over `map_err` otherwise you are dropping the error history.
2. `with_context` is being called right before this, than can go away now.

thomaseizinger (changes requested):
Neat, thanks for bringing in http-api-problem!

Two comments regarding the use of anyhow.

thomaseizinger (approved):

luckysori (approved):

PR #102, opened 2021-01-06
Display decoded transaction data

### LINKED ISSUE #95: Show swap data based on decomposed transaction
Basically this TODO here: https://github.com/thomaseizinger/droplet/blob/8b65bd1dc40818cb9e16cfc27f70f3a8baea3744/waves/src/ConfirmSwapDrawer.tsx#L45-L49

luckysori:
I decided not to integrate with nigiri-chopsticks' or esplora's registry for the purpose of identifying the assets in the transaction because I was struggling to make it work (esplora's returns multiple pages, nigiri-chopsticks' has a different format (without precision), talking to nigiri-chopsticks from the frontend was not working). I'm almost certain it is not worth integrating with nigiri-chopsticks, but we may want to detect when we're in production and figure out how to consume esplora's registry instead of depending on the local registry.

My React is very rusty, so there's probably much nicer ways of doing the same thing.

I am not too happy with the number of things that are being done by the frontend, so I would consider moving almost all the logic into the wallet and changing the API to be something very specific to what we need.

Resolves #95


waves-scripts
```diff
@@ -82,9 +82,10 @@ start_bobtimus () {
 
     set +e
     find ./waves/dist ./waves/dist/index.html &> /dev/null
+    [ $? -ne 0 ] && is_built=false
```
bonomat, line 85: Nitpick: Double negations are hard to read.

How about `build_failed=true` and then below you say `if build_failed; then`...

waves/wallet/src/wallet/decompose_transaction.rs
```diff
@@ -67,8 +67,27 @@ pub async fn decompose_transaction(
         })
         .collect::<Result<Vec<_>, UnblindError>>())?;
 
-    Ok(TransactionElements {
```
bonomat, line 70: Why did you opt for removing `TransactionElements` in the favor of one amount being negative and one positive?

bonomat (changes requested):
I don't really like the approach of negative amounts and then _re-negating_ it again in the frontend. I'm afraid that this complexity will bite us in the future. At the very least it should be documented on the function `decompose_transaction` so that a caller knows what to expect. 

What was the particular reason you opted for removing `TransactionElements` and using negative amounts?

waves-scripts
```diff
@@ -82,9 +82,10 @@ start_bobtimus () {
...
     set -e
 
-    if [ $? -ne 0 ]; then
+    if ! $is_built; then
```
thomaseizinger, line 88: I think we should not be implicitly building the frontend here because we are not serving it through bobtimus in development anyway.

waves/src/ConfirmSwapDrawer.tsx
```diff
@@ -29,24 +29,78 @@ interface ConfirmSwapDrawerProps {
...
+            transactionBalance.forEach(([ticker, amount]) => {
+                if (amount < 0) {
+                    let asset = tickerToAssetType(ticker);
+                    let balanceBefore = asset === AssetType.BTC ? btcBalance : usdtBalance;
+                    let balanceAfter = +balanceBefore + +amount;
+
+                    setAlphaAmount(-amount);
+                    setAlphaAsset(asset);
+                    setAlphaBalanceBefore(balanceBefore);
+                    setAlphaBalanceAfter(balanceAfter);
```
thomaseizinger, line 82: I _think_ having several `set`'s is not really good for performance because it triggers multiple re-renders if React fails to batch them together.

waves/src/ConfirmSwapDrawer.tsx
```diff
@@ -29,24 +29,78 @@ interface ConfirmSwapDrawerProps {
...
+            case LBTC_TICKER:
+                return AssetType.BTC;
+            case LUSDT_TICKER:
+                return AssetType.USDT;
+            default:
+                throw new Error("Unexpected ticker: " + ticker);
+        }
+    }
+
+    useAsyncEffect(async () => {
```
thomaseizinger, line 59: Instead of using `useAsyncEffect` and several `useStates` how about:

- 1 `useAsync` with `promiseFn` (that will make it execute immediately instead of having to trigger it with `run`)
- You can use the `watch` property to re-execute if `transaction` changes
- You can use the `isPending` property to check if the computation is done and display a loading spinner where you would normally display the balances while it is computing
- You can also chain all the functions like `tickerToAssetType` onto the `decomposeTransaction` like this:

```ts
useAsync({
	promiseFn: async () => {
		let transactionBalance = await decomposeTransaction(transaction);

		// do all your things here

		return {
			alphaAsset: ...,
			alphaAmount: ...,
			...
		}
	}
})
```

waves/wallet/src/registry.rs
```diff
@@ -0,0 +1,13 @@
+use elements_fun::AssetId;
+use std::collections::HashMap;
+
+pub fn registry() -> HashMap<AssetId, (String, u8)> {
```
thomaseizinger, line 4: Could this function maybe be `const`?

You might need to change the `String` for `&'static str`.

thomaseizinger: Also, it might be better to expose a lookup function from this module instead of the hashmap (which is kind of an implementation detail).

thomaseizinger (changes requested):
ACK for using constants for the two assets we care about.

Made some suggestions for how to optimize the TS code.

thomaseizinger (approved):

bonomat (approved):
LGTM

PR #110, opened 2021-01-07
Move new/unlock wallet to the top and disable swap button

### LINKED ISSUE #104: Make wallet more discoverable
- Introduce "Connect" button at the same location as balances if wallet is not available or locked
- Center button always says "Swap", is greyed out if wallet is locked / not available

### LINKED ISSUE #105: Disable swap button if balance lower than swap amount
Otherwise it blows up.

bonomat:
Resolves #105.
Resolves #104.

Unlock wallet: 
image

Swap Button disabled: 
image

### DIFF
--- a/waves/src/App.tsx
+++ b/waves/src/App.tsx
@@ -245,11 +245,15 @@ function App() {
             Unlock wallet
         </Button>;
     } else {
+        let disabled = state.alpha.type == AssetType.BTC
+            ? btcBalance < state.alpha.amount
+            : usdtBalance < state.alpha.amount;
         button = <Button
             onClick={makeNewSwap}
             size="lg"
             variant="main_button"
             isLoading={isCreatingNewSwap}
+            disabled={disabled}
         >
             Swap
         </Button>;
--- a/waves/src/App.tsx
+++ b/waves/src/App.tsx
@@ -224,50 +224,43 @@ function App() {
         rate,
     );
 
-    let button;
+    let walletInfoButton;
 
     if (!walletStatus.exists) {
-        button = <Button
+        walletInfoButton = <Button
             onClick={onCreateWalletOpen}
-            size="lg"
-            variant="main_button"
+            size="sm"
+            variant="connect_wallet_button"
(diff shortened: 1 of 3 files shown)


waves/src/theme.tsx
```diff
@@ -50,6 +42,16 @@ const theme = extendTheme({
                     color: "white",
                     _hover: {
                         bg: "blue.300",
+                        _disabled: {
```
bonomat, line 45: We had a _bug_ where the button disappeared when disabled and hovered over. Best would probably be to disable hovering if the button is disabled but I couldn't figure out how.

waves/src/App.tsx
```diff
@@ -302,7 +297,15 @@ function App() {
                                 <Text textStyle="info">1 BTC ~ {rate.ask} USDT</Text>
                             </Box>
                             <Box>
-                                {button}
+                                <Button
+                                    onClick={makeNewSwap}
+                                    size="lg"
+                                    variant="main_button"
+                                    isLoading={isCreatingNewSwap}
+                                    disabled={disableSwapButton}
```
thomaseizinger, line 305: I'd prefer naming of these variables in terms of states instead of actions, e.g. `isSwapButtonDisabled`.

waves/src/theme.tsx
thomaseizinger: Any reason why we are even overriding the `hover` styles?

To my knowledge, with all the styling in general, it is usually more consistent to modify the theme and stick to the pre-configured variants instead of inventing your own variants.

thomaseizinger (approved):

waves/src/theme.tsx
bonomat: It could be that I'm using the styles wrongly. The issue is that the default color is `grey` and hovering makes it darker. 

I wanted to change the background color (`bg`) to blue and that's where this weird behavior showed up, i.e. I had to set `hover` to a color as well, otherwise no hovering would happen and when the button is disabled the hovering changed the button color to white.

PR #112, opened 2021-01-07
Use debug instead of console.log

### LINKED ISSUE #107: Frontend: use `debug` instead of console.{log,warn, ...}
https://github.com/visionmedia/debug

bonomat:
When starting the application we assert if the application was built in development or in production mode. If in production we give a warning indicating that the user won't see any logs if `debug` is not enabled. If in development mode, we set debug variable to `*` to enable all logs.

Note: if you run the application in chrome you will need to enable `verbose` log level as described here: https://github.com/visionmedia/debug/issues/803
Otherwise you won't see any logs. 

Resolves #107


luckysori (approved):
:star2:

thomaseizinger (approved):

PR #114, opened 2021-01-08
Stop making too many evil clones of Bob

### LINKED ISSUE #91: Bobtimus seems to not be able to handle more than one swap
Getting this error when trying to do a 2nd swap: 

```
[2021-01-06T00:32:40Z ERROR bobtimus::http] Failed to handle create swap: failed to prove surjection
```

luckysori:
Bobtimus holds an instance of rng initialised at start-up. Every time we receive a request, we clone our instance of Bobtimus and therefore the instance of rng itself. Any state changes to the rng (which happen during usage) thereby did not affect the original prototype of Bobtimus. As a result, any future request started with exactly the same rng state and therefore produced the same sequence of random numbers.

This led to conflicting blinding factors between inputs and outputs.

Fixes #91.

### DIFF
--- a/bobtimus/src/bin/bobtimus.rs
+++ b/bobtimus/src/bin/bobtimus.rs
@@ -5,7 +5,9 @@ use elements_fun::{
     secp256k1::rand::{rngs::StdRng, thread_rng, SeedableRng},
 };
 use elements_harness::Client;
+use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):

bonomat (commented):
Good job.

bonomat (approved):

PR #115, opened 2021-01-08
Fix decimal number input issue.

### LINKED ISSUE #103: It's not possible to type decimal numbers into the amount field
Once deleted, one can only sell a fraction of a BTC by using the arrow keys.

bonomat:
The problem was that the component `NumberInput` propagates the value as a string or as a number up onChange. Since we dropped/ignored the string any added decimal points were overwritten.

Resolves #103


thomaseizinger (approved):
Nice!

PR #116, opened 2021-01-08
Improve error handling and make log level configurable

### LINKED ISSUE #72: Wallet: Initialize logger only if specific variable is set in local storage
- Listen on localstorage key `wallet_log`
- Parse value as log level
- No key means off

thomaseizinger:
Fix #72.


waves/wallet/src/logger.rs
```diff
@@ -0,0 +1,16 @@
+use crate::storage::Storage;
+use anyhow::{Context, Result};
+
+pub fn try_init() -> Result<()> {
+    let local_storage = Storage::local_storage()?;
+
+    let log_level = local_storage
+        .get_item::<log::Level>("wallet_log")
+        .context("failed to get `wallet_log` log level")?;
+
```
bonomat, line 10: Will the user see an error message if `wallet_log` is not set in localStorage?

bonomat (approved):
Just one question, otherwise looks good.

waves/wallet/src/logger.rs
thomaseizinger: Nope, by default the console will be quiet.

However, I don't think it is an _error_ to not set it. The logger is a developer / debugger setting so the ordinary user should not need to concern themselves with it.

Anyone interested in hacking on this thing can always read the documentation (README.md) to learn how to enable the logs.

luckysori (approved):
I love 002c721858230d566642aae3a8fc150ccf98ed84

thomaseizinger:
> I love [002c721](https://github.com/comit-network/droplet/commit/002c721858230d566642aae3a8fc150ccf98ed84)

I don't understand how I didn't think of that before! :)

PR #119, opened 2021-01-08
Serve frontend under /app namespace

### LINKED ISSUE #79: Consider serving waves frontend under its own namespace
Rationale: future-proof against route-conflicts between api, assets and frontend routes

- adapt warp to serve assets (excluding index.html) under namespace
- re-configure webpack to generate appropriate relative links (likely a package.json config)
- actual name of namespace is not that important (i.e. choose something meaningful)

thomaseizinger:
This applies to all assets apart from index.html.

Unfortunately, we need to add a hook to the app to redirect from
/app to / to make it work with the webpack-dev-server because
setting `homepage` in `package.json` makes `react-scripts` open
the browser on `http://localhost:3004/app`.

I hesitated to guard this piece of code with `if env == development`
because I think it might actually also be useful in production.
If someone for whatever reasons investigates our routes and finds
`/app`, we just send them back to the root.

Resolves #79.

### DIFF
--- a/bobtimus/src/http.rs
+++ b/bobtimus/src/http.rs
@@ -24,11 +24,14 @@ where
     R: RngCore + CryptoRng + Clone + Send + Sync + 'static,
     RS: LatestRate + Clone + Send + Sync + 'static,
 {
-    let index_html = warp::path::end().and_then(serve_index);
-    let waves_resources = warp::path::tail().and_then(serve_waves_resources);
+    let index_html = warp::get().and(warp::path::end()).and_then(serve_index);
+    let waves_resources = warp::get()
+        .and(warp::path("app"))
+        .and(warp::path::tail())
+        .and_then(serve_waves_resources);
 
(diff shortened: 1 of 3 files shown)


bonomat (approved):
LGTM

bobtimus/src/http.rs
```diff
@@ -24,11 +24,14 @@ where
     R: RngCore + CryptoRng + Clone + Send + Sync + 'static,
     RS: LatestRate + Clone + Send + Sync + 'static,
 {
-    let index_html = warp::path::end().and_then(serve_index);
-    let waves_resources = warp::path::tail().and_then(serve_waves_resources);
+    let index_html = warp::get().and(warp::path::end()).and_then(serve_index);
```
luckysori, line 27: :question: Was the `GET` implicit before?

luckysori (approved):

bobtimus/src/http.rs
thomaseizinger: I guess so :)

PR #127, opened 2021-01-12
Use secp256k1-zkp from crates.io

thomaseizinger:


bonomat (approved):
🥳

Regarding the test failure: I triggered a restart. The issue was that the SwapButton was not enabled, i.e. we didn't see the funding. I've created an issue for that: #128 .

PR #133, opened 2021-02-05
Reduce custom styling

### LINKED ISSUE #122: Reduce custom styling
We currently using several custom styles for buttons and other things by introducing "variants" in the theme:

https://github.com/comit-network/droplet/blob/85094e61af25d6894803e8eaad45b90bc9e06cee/waves/src/theme.tsx#L34-L63

This easily leads to inconsistent usage of colors and doesn't scale.

We should pick a theme (i.e. a color palette) to our liking and use the built-in variants of buttons instead.

thomaseizinger:
Fixes #122.

I chose blue as our primary color because water is blue and the project is called waves.

Some screenshots:

image

image

image

### DIFF
--- a/waves/src/theme.tsx
+++ b/waves/src/theme.tsx
@@ -2,20 +2,16 @@ import { extendTheme } from "@chakra-ui/react";
 
 const theme = extendTheme({
     textStyles: {
-        actionable: {
-            fontSize: "lg",
-            color: "gray.500",
-        },
-        info: {
+        smGray: {
             fontSize: "sm",
             color: "gray.500",
         },
-        addressInfo: {
-            fontSize: "sm",
-            color: "gray.500",
-        },
-        assetSelect: {
+        mdGray: {
             fontSize: "md",
+            color: "gray.500",
+        },
+        lgGray: {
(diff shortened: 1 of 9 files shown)


waves/src/theme.tsx
```diff
@@ -17,44 +17,21 @@ const theme = extendTheme({
...
-                "main_button": {
-                    w: "15rem",
+                "primary": {
                     color: "white",
+                    bg: "blue.500",
                     _hover: {
                         bg: "blue.300",
                         _disabled: {
-                            bg: "",
+                            bg: "blue.300",
```
bonomat, line 27: Did you try if this works as expected? 

If the button is `disabled` it has a different color (usually brighter) and we used to have a weird bug here where hovering over a disabled button caused the button to disappear (turn almost completely white).

thomaseizinger: |Disabled|Hover & Disabled|
|---|---|
|image|image|

bonomat: cool 👍

luckysori (approved):
:blue_heart:

PR #134, opened 2021-02-05
Replace elements-fun with rust-elements

luckysori:
Introduces a patch dependency on `luckysori/rust-elements` until https://github.com/ElementsProject/rust-elements/pull/70 is merged.


thomaseizinger (commented):
You will need to remove the fuzz CI job :)

waves/wallet/build.rs
```diff
@@ -41,16 +41,18 @@ fn main() -> Result<()> {
         &constants_rs,
         &format!(
             r#"
+use conquer_once::Lazy;
+
 pub const NATIVE_ASSET_TICKER: &str = "{}";
-pub const NATIVE_ASSET_ID: elements_fun::AssetId = elements_fun::AssetId::from_bytes({:?});
-pub const USDT_ASSET_ID: elements_fun::AssetId = elements_fun::AssetId::from_bytes({:?});
+pub static NATIVE_ASSET_ID: Lazy<elements::AssetId> = Lazy::new(|| elements::AssetId::from_slice(&{:?}).unwrap());
```
thomaseizinger, line 47: Record an issue for this please.

waves/wallet/src/typed_js_future.rs
```diff
@@ -1,4 +1,4 @@
-use elements_fun::bitcoin_hashes::core::pin::Pin;
+use bitcoin_hashes::core::pin::Pin;
```
thomaseizinger, line 1: That should have been an `std` import in the first place :)

luckysori:
> You will need to remove the fuzz CI job :)

Quick catch, thanks!

waves/wallet/build.rs
luckysori: https://github.com/comit-network/waves/issues/135

.github/workflows/ci.yml
```diff
@@ -148,9 +108,9 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: cargo test --workspace --all-features
 
-      - name: Cargo build (all workspace without elements-fuzz )
+      - name: Cargo build
         if: matrix.os == 'macos-latest'
-        run: cargo build --workspace --exclude elements-fuzz --all-features
+        run: cargo build --workspace --all-features
```
thomaseizinger, line 113: This can be removed now together with line 108.

waves/wallet/src/wallet.rs
```diff
@@ -308,8 +306,17 @@ fn compute_balances(wallet: &Wallet, txouts: &[TxOut]) -> Vec<BalanceEntry> {
...
         .filter_map(|utxo| match utxo {
-            TxOut::Explicit(explicit) => Some((explicit.asset.0, explicit.value.0)),
-            TxOut::Confidential(confidential) => {
+            TxOut {
+                asset: confidential::Asset::Explicit(asset),
+                value: confidential::Value::Explicit(value),
+                ..
+            } => Some((*asset, *value)),
+            txout => {
+                let confidential = match txout.clone().into_confidential() {
```
thomaseizinger, line 315: I see you are calling `.clone().into` a lot. That sounds like you'd want `to_confidential` as well :)

luckysori: I think you're referring to `as_confidential`. I tried to add it to `rust_elements`, but I failed.

thomaseizinger: Not quite. `as_confidential` would imply as signature of `&self -> &T`. That doesn't work because we no longer store an instance of `ConfidentialTxOut` inside `TxOut` but construct it on the fly instead which makes it impossible to return a reference (what would the reference refer to?).

My suggested `to_confidential` would be similar to `into_confidential` but take `&self` instead of `self`. As per Rust's naming convention `to_` implies an allocation which is what we would need to do for `&self -> T`.

luckysori: That makes sense. I guess it's preferable to do all the allocations in a single place, as long as the name of the method indicates what's going on.

thomaseizinger: Also, nothing prevents us from having `into_` and `to_` :)

.github/workflows/ci.yml
luckysori: If I remove what you suggest we run docker tests on MacOS which [fails](https://github.com/comit-network/waves/pull/134/checks?check_run_id=1859103372#step:5:492). Did I misunderstand you?

thomaseizinger: Ah yeah, I am an idiot. Mixed up build and test script :(

elements-harness/src/elementd_rpc.rs
```diff
@@ -196,10 +200,10 @@ impl Client {
                 .await
                 .context("cannot unblind raw source transaction")?;
             if unblinded_raw_tx.output[source_vout as usize]
-                .as_explicit()
+                .clone()
+                .into_explicit()
```
thomaseizinger, line 204: I see you added `to_confidential` in the upstream PR. Here is a piece of code where you could take advantage of `to_explicit`.

swap/tests/happy_path.rs
```diff
@@ -216,7 +216,7 @@ async fn move_output_to_wallet(
     let previous_output = previous_output_tx.output[previous_output.vout as usize].clone();
 
     let txout = previous_output
-        .as_confidential()
+        .to_confidential()
         .context("not a confidential txout")?
         .clone();
```
thomaseizinger, line 221: Why is this `clone` necessary?

waves/wallet/src/wallet/coin_selection.rs
```diff
@@ -141,7 +142,7 @@ fn max_satisfaction_weight(script_pubkey: &Script) -> Option<usize> {
 mod tests {
     use std::str::FromStr;
 
-    use elements_fun::{Address, Txid};
+    use elements::{Address, Txid};
```
thomaseizinger, line 146: We seem to have some unnecessary newlines here.

thomaseizinger (approved):
Good to go in from my side! Thank you for making this effort @luckysori !

elements-harness/src/elementd_rpc.rs
luckysori: Ahhhh, I forgot to add `to_explicit`, what a twit.

swap/tests/happy_path.rs
luckysori: Nice, missed that one!

PR #136, opened 2021-02-15
Let Alice buy bitcoin

luckysori:


bobtimus/src/lib.rs
```diff
@@ -395,6 +420,141 @@ mod tests {
...
+        // let address = client.getnewaddress().await.unwrap();
+        // let _txid = client
+        //     .send_asset_to_address(
+        //         &address,
+        //         Amount::from_btc(10.0).unwrap(),
+        //         Some(have_asset_id_bob),
+        //     )
+        //     .await
+        //     .unwrap();
+        // client.generatetoaddress(1, &mining_address).await.unwrap();
```
bonomat, line 484: this can probably be removed.

bobtimus/src/lib.rs
```diff
@@ -395,6 +420,141 @@ mod tests {
...
+                None,
+                Some(ListUnspentOptions {
+                    asset: Some(have_asset_id_alice),
+                    ..Default::default()
+                }),
+            )
+            .await
+            .unwrap();
+
+        let error = 0.0001;
```
bonomat: What's the error for?

bobtimus/src/lib.rs
```diff
@@ -395,6 +420,141 @@ mod tests {
...
+
+        let have_asset_id_alice = client.issueasset(100_000.0, 0.0, true).await.unwrap().asset;
+        let have_asset_id_bob = client.get_bitcoin_asset_id().await.unwrap();
+
+        let mut rate_service = fixed_rate::Service::new();
+        let sell_amount_bob = LiquidBtc::from(Amount::ONE_BTC);
+        let redeem_amount_bob = rate_service
+            .latest_rate()
+            .sell_quote(sell_amount_bob)
+            .unwrap();
```
bonomat: To make it more explicit I would type out the fixed amounts here.

bobtimus/src/lib.rs
```diff
@@ -395,6 +420,141 @@ mod tests {
...
+                }],
+                address: final_address_alice,
+                btc_amount: sell_amount_bob,
+            })
+            .await
+            .unwrap();
+
+        let transaction = swap::alice_finalize_transaction(transaction, {
+            let commitment = input_alice.1.into_confidential().unwrap().value;
+            move |mut tx| async move {
```
bonomat: Does this have to be async?

waves/src/App.tsx
```diff
@@ -297,6 +311,7 @@ function App() {
                                                         betaAmount,
                                                     },
                                                 })}
+                                            dataCy="exchange-asset-types"
```
bonomat, line 314: Let's have the naming convention 
`xx-button` if it is a button,
`xx-input` if it is an input field, 
`xx-textfield` if it is a textfield (read only). 

```suggestion
                                            dataCy="exchange-asset-types-button"
```

waves/wallet/src/wallet/make_create_swap_payload.rs
```diff
@@ -52,13 +88,26 @@ pub async fn make_create_sell_swap_payload(
...
-    let fee_offset = calculate_fee_offset(bobs_fee_rate);
+    let (bobs_fee_rate, fee_offset) = if fee_asset == sell_asset {
+        // Bob currently hardcodes a fee-rate of 1 sat / vbyte, hence
+        // there is no need for us to perform fee estimation. Later
+        // on, both parties should probably agree on a block-target
+        // and use the same estimation service.
+        let bobs_fee_rate = Amount::from_sat(1);
+        let fee_offset = calculate_fee_offset(bobs_fee_rate);
+
+        (bobs_fee_rate, fee_offset)
```
bonomat, line 99: I just realized that we might have a very small issue: 
If Alice sells BTC we check on the UI if her `sell_amount < balance`. 
If I see it correctly, then we add the `fee_offset` on top of `sell_amount` which means the coin selection algorithm might fail if she tries to sell her full balance.

bonomat (approved):
Good work and very clean commit history 👍 . I didn't check why the tests fail.

bobtimus/src/lib.rs
bonomat: Resolved: saw this was fixed in a later commit.

bobtimus/src/lib.rs
luckysori: It was to account for the fee. But good thing you asked, because Alice pays for the fee in this case so Bob actually gets the expected `redeem_amount_bob` exactly.

Furthermore, for the other test we can use `transaction.fee_in` to extract the exact fee paid in bitcoin and assert value equality, rather than using an error estimate.

bobtimus/src/lib.rs
luckysori: Yep, the second argument (`S`) to this function has to fulfil
```
S: FnOnce(Transaction) -> F,
F: Future<Output = Result<Transaction>>,
```
so this closure must return a future.

waves/wallet/src/wallet/make_create_swap_payload.rs
luckysori: Yeah, you're right. I'll create an issue!

luckysori: Recorded here: https://github.com/comit-network/waves/issues/137.

PR #140, opened 2021-03-08
Extract wallet into web extension

luckysori:

PR #148, opened 2021-03-14
Extract the wallet from waves into a browser extension.

bonomat:
This commit moves the wallet out of the `waves` module and introduces a new module `extension`. This extension consists out of a few sub-modules/scripts:

* background script: contains the core logic of the wallet. Found in extension/background.
* popup script: the logic of the extension's frontend. Found in extension/src.
* content script: injects an in-page script into the webpage and acts as a thin communication layer between the in-page script and the background script. This communication is based on events.
* in-page script: provides proxy functions to interact with the wallet (background script).

Additionally, the end2end tests are now using selenium because Cypress does not support browser extensions. The end2end tests can be found in `e2e_tests.ts` which also contains its own setup script `e2e_test_setup.sh`.

Co-authored-by: Lucas Soriano del Pino <lucas@coblox.tech>

original branch with full commit history can be found here: https://github.com/comit-network/waves/pull/140

supersedes #140


waves/src/App.test.tsx
```diff
@@ -17,20 +17,6 @@ class DummySource implements Source {
     }
 }
 
-test("if rendering works by asserting `create new wallet` button", () => {
-    act(() => {
-        render(
-            <SSEProvider source={() => new DummySource()}>
```
bonomat, line 23: I guess you can also delete `DummySource` a few lines up.

luckysori (approved):

PR #151, opened 2021-03-15
Update readme

luckysori:


waves-scripts
```diff
@@ -107,36 +121,44 @@ start_bobtimus () {
...
     __read_env_file
 
     cd waves
     yarn install
     export REACT_APP_BLOCKEXPLORER_URL="http://localhost:$LIQUID_ESPLORA_PORT"
 
-    yarn start
+    yarn run build
+
+    cd ..
```
bonomat, line 139: Did you test that this works? 
In `start_webapp` you call `build_webapp` which changes directory into `waves` and then out again but afterwards on line 125 you call `yarn run start`?

bonomat (approved):
Looks good. Please double check that the script works.

waves-scripts
luckysori: Good catch! I was only testing `start_extension_dev_env`, so I missed this. Thank you.

PR #152, opened 2021-03-17
Add liquid field to window object instead of methods

luckysori:
For this to work we must use `&JsValue::from(liquid)` and not `&JsValue::from_serde(&liquid)` (see https://github.com/rustwasm/wasm-bindgen/issues/2493).


bonomat (approved):
Awesome. Well done 🥳

PR #153, opened 2021-03-23
Loan protocol

luckysori:


.github/workflows/ci.yml
```diff
@@ -216,3 +216,19 @@ jobs:
...
+
+      - name: Install Rust toolchain
+        id: toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          override: true
+
+      - name: Cargo test
+        run: cargo test -p covenants
```
luckysori, line 234: Ooops, we don't actually need this. There already was a job executing cargo tests for the whole workspace.

bonomat: True that :)

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+                        && out.script_pubkey == self.address.script_pubkey();
+
+                    predicate.then(|| out)
+                }
+                None => None,
+            })
+            .context("no collateral change txout")?;
+
+        Ok(Borrower1 {
+            keypair: self.keypair,
```
bonomat, line 211: nit: renaming to what this key is used.

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+            tx.input[1].witness = TxInWitness {
+                amount_rangeproof: vec![],
+                inflation_keys_rangeproof: vec![],
+                script_witness,
+                pegin_witness: vec![],
+            };
+        };
+
+        // TODO: Sign more than one input if necessary
+        // sign repayment input of the principal amount
```
bonomat, line 444: No need to do this now, but once extracted into a lib we should do this!

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+impl Lender0 {
+    pub fn new<C>(
+        secp: &Secp256k1<C>,
+        bitcoin_asset_id: AssetId,
+        usdt_asset_id: AssetId,
+        // TODO: Here we assume that the wallet is giving us _all_ the
+        // inputs available. It would be better to coin-select these
+        // as soon as we know the principal amount after receiving the
+        // loan request
+        principal_inputs: Vec<Input>,
```
bonomat, line 477: We should pass in a coin-selector as in the other function e.g. see `borrower.loan_repayment_transaction(..)`.

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+                collateral_change_tx_out,
+                tx_fee_tx_out,
+            ],
+        };
+
+        let repayment_collateral_input = Input {
+            tx_in: TxIn {
+                previous_output: OutPoint {
+                    txid: loan_transaction.txid(),
+                    vout: 0,
```
bonomat, line 684: This looks a bit dangerous to me. 😬

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+    lender_pk: PublicKey,
+    repayment_collateral_input: Input,
+    repayment_collateral_abf: AssetBlindingFactor,
+    repayment_collateral_vbf: ValueBlindingFactor,
+    timelock: u64,
+    repayment_principal_output: TxOut,
+}
+
+pub struct Borrower0 {
+    keypair: (SecretKey, PublicKey),
```
bonomat, line 70: nit: do we need to keep the `PublicKey` around?

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+        // input specific values
+        let (previous_out, script_0, script_1, script_2, value, sequence) = {
+            let InputData {
+                previous_output,
+                script,
+                value,
+                sequence,
+            } = &self.input;
+
+            let third = script.len() / 3;
```
bonomat, line 1007: We should fail here if `script.len()/3 > 80`

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+
+            if self.other_outputs.len() > 3 {
+                bail!("too many outputs");
+            }
+
+            for txout in self.other_outputs.iter() {
+                let mut output = Vec::new();
+                txout.consensus_encode(&mut output)?;
+
+                let middle = output.len() / 2;
```
bonomat, line 1059: Same as above, fail if `output.len()/2>80`

covenants/src/protocol_tests.rs
```diff
@@ -0,0 +1,403 @@
...
+    use elements_harness::{elementd_rpc::ElementsRpc, Elementsd};
+    use secp256k1_zkp::Message;
+    use std::env;
+    use testcontainers::clients::Cli;
+
+    #[tokio::test]
+    async fn borrow_and_repay() {
+        // force enabling log output
+        // TODO: remove me again
+        env::set_var("RUST_LOG", "DEBUG");
```
bonomat, line 24: Can be removed now.

covenants/src/protocol_tests.rs
```diff
@@ -0,0 +1,403 @@
...
+                timelock,
+                bitcoin_asset_id,
+                usdt_asset_id,
+            )
+            .unwrap();
+
+            (borrower, address)
+        };
+
+        // TODO: Use a separate wallet per actor. Using the same wallet is confusing and bug-prone.
```
bonomat, line 86: What do you think @luckysori : shall we tackle this before merging? 
Could be easier to find potential flaws now than later.

covenants/src/protocol_tests.rs
```diff
@@ -0,0 +1,403 @@
...
+        let (lender, _lender_address) = {
+            let address = client
+                .get_new_address(Some("blech32".into()))
```
bonomat, line 89: we could add a helper function to return a confidential address which is in blech32 format, e.g. 
`client.get_new_confidential_address()`.

covenants/src/lib.rs
```diff
@@ -0,0 +1,1160 @@
...
+                .encode_segwitv0_signing_data_to(
+                    &mut unhashed,
+                    1,
+                    &self.collateral_script.clone(),
+                    self.repayment_collateral_input.original_tx_out.value,
+                    SigHashType::All,
+                )
+                .unwrap();
+
+            dbg!(hex::encode(unhashed));
```
bonomat, line 406: Can be removed now.

covenants/src/protocol_tests.rs
```diff
@@ -0,0 +1,403 @@
...
+            .await
+            .unwrap();
+
+        client
+            .send_raw_transaction(&loan_repayment_transaction)
+            .await
+            .unwrap();
+    }
+
+    // #[tokio::test]
```
bonomat, line 194: Ups, completely forgot about this :D

covenants/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+
+        let sighash_type = {
+            let mut writer = Vec::new();
+            self.sighash_type.as_u32().consensus_encode(&mut writer)?;
+            writer
+        };
+
+        Ok(vec![
+            sig,
+            pk,
```
luckysori, line 1069: @bonomat remind me: why do we need this pk in the witness stack? Don't we already have it in the script? Is it just for convenience?

covenants/src/lib.rs
luckysori: What did you have in mind? Isn't this just the borrower's keypair?

covenants/src/lib.rs
luckysori: We use it both the lender's and borrower's PKs every time we compute the loan contract, so I think we do.

covenants/src/protocol_tests.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        client
+            .send_raw_transaction(&loan_repayment_transaction)
+            .await
+            .unwrap();
+    }
+
+    fn init_logger() {
+        // force enabling log output
+        // TODO: remove me again
```
luckysori, line 193: :question: @bonomat So do you want to remove this helper function or keep it?

covenants/src/lib.rs
bonomat: Yes, but for what? I honestly forgot for what atm :D

covenants/src/protocol_tests.rs
bonomat: Keep the `init_logger` function but remove the next line: `env::set_var("RUST_LOG", "DEBUG");` which should be set from the outside

covenants/src/lib.rs
bonomat: You are right, we have it already in the script `        .push_slice(&borrower_pk.serialize())
`. 
I don't think there was a particular reason for having it here again. IF we change it, we will need to make use of `OP_DUP` and move the `pk` in the script around a couple of times or add it to the script twice. (we need it for `op_checksig` and `op_checksigfromstack`).

luckysori: See #155

PR #158, opened 2021-05-25
Move faucet into webapp

bonomat:
It makes more sense to have the faucet in the webapp than in the browser extension as it is a _feature_ by bobtimus.


luckysori (approved):
Can we make it so that the button is not rendered in production? Is it already the case?

luckysori:
Dang, mergify is too speedy for me.

PR #162, opened 2021-06-07
Update to later version of `rust-elements`

luckysori:
There's still a smaller patch dependency which allows us to unblind transaction outputs. Once https://github.com/ElementsProject/rust-elements/pull/78 is merged, we can remove it.

We expect `rust-elements` to support PSETs fully in the near future, which should allow us to simplify a lot of this code anyway. But updating the code was simple enough.


luckysori:
Superseded by #163.

PR #163, opened 2021-06-10
Rebase with 'loan-tab' branch

DeliciousHair:
Intermediate action to prevent `loan-tab` from diverging from `update-rust-elements` branch. Suggest closing `loan-tab` branch.

PR #164, opened 2021-06-17
Add db support to bobtimus

bonomat:
Some notes: 

- to streamline with our other projects I opted for using tracing instead of env_logger. 
- atm bobtimus accepts runtime arguments for the settings. eventually we want this to be a config file. I did a small refactoring commit which goes into this direction. This simplified the use of default values for the db file.
- the third commit adds db support to bobtimus :) 
- the last commit is just a demo showing how the db can be used in a post request. 
I added this demo to test the mutex behaviour around the db. 
This proofs that it "works" and we can store data in the db from within a post request :) 
Once approved I'll drop this commit, so no need to review it. 

Once merged, we should rebase https://github.com/comit-network/waves/pull/163 and we can start using the db.


bonomat:
Reverted to draft so that it does not accidentally get merged before I can drop the last commit :)

luckysori (approved):

PR #167, opened 2021-07-14
Mac os extension build fix

rishflab:


README.md
```diff
@@ -11,7 +11,6 @@ This project includes:
 
 ## Try it out on _regtest_
 
-
```
rishflab, line 14: dprint did this

README.md
```diff
@@ -68,6 +70,27 @@ yarn install # only needed once
...
+
+```bash
+brew install llvm
+```
+
+Run the extension
+
+```bash
+cd extension;
+CC=/usr/local/opt/llvm/bin/clang AR=/usr/local/opt/llvm/bin/llvm-ar yarn install
```
da-kami, line 88: yarn install should not be concerned with these envvars

rishflab: are you suggesting that the setting of env vars be done a different line?

README.md
```diff
@@ -68,6 +70,27 @@ yarn install # only needed once
...
+CC=/usr/local/opt/llvm/bin/clang AR=/usr/local/opt/llvm/bin/llvm-ar yarn watch
+```
+
+## Running the Web App
```
da-kami, line 99: This chapter is confusing and should be removed.

Maybe we should generally not introduce these chapters. I propose to mention the MacOs setup inline or at the bottom. 

yarn start is part of `Run the Web Extension` - it is the command that actually starts a browser instance that has the web extension installed.

README.md
da-kami: I am saying, that only the `yarn watch` is concerned with the envvars `CC` and `AR`, because `yarn watch` triggers the wasm compilation. `yarn install` only installs nodejs packages and is not concerned with wasm.

README.md
rishflab: Ill get rid of this chapter but I think we still need the other ones. It doesn't seem understandable otherwise

da-kami: I pushed a few fixes including one that fixes the overall chapter structure.

bonomat (approved):

PR #168, opened 2021-07-14
simplified local test

DeliciousHair:
create a simple if-then condition to avoid having the user remembering the correct way to run the e2e tests locally

### DIFF
--- a/e2e_tests/e2e_test_setup.sh
+++ b/e2e_tests/e2e_test_setup.sh
@@ -1,8 +1,15 @@
 set -e
 
-# This script is for CI purposes only. If you want to run the e2e tests locally make sure to setup your environment accordingly using:
-#          curl https://travis.nigiri.network | bash
-#          docker-compose -f ./docker-compose.yml up -d
+# This script is primarily for CI purposes. If you want to run the e2e tests locally use the -L flag:
+#    ./e22_test_setup.sh -L
+
+if getopts ":L" arg; then
+  mkdir nigiri
+  cd nigiri
+  curl https://travis.nigiri.network | bash
+  docker-compose -f ./docker-compose.yml up -d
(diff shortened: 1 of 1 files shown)


e2e_tests/e2e_test_setup.sh
```diff
@@ -1,8 +1,15 @@
 set -e
 
-# This script is for CI purposes only. If you want to run the e2e tests locally make sure to setup your environment accordingly using:
-#          curl https://travis.nigiri.network | bash
-#          docker-compose -f ./docker-compose.yml up -d
+# This script is primarily for CI purposes. If you want to run the e2e tests locally use the -L flag:
+#    ./e22_test_setup.sh -L
+
+if getopts ":L" arg; then
+  mkdir nigiri
```
bonomat, line 7: This will fail if it already exists right?

DeliciousHair: You're correct, oops! Lemme adjust that.

e2e_tests/e2e_test_setup.sh
```diff
@@ -1,8 +1,17 @@
...
-#          docker-compose -f ./docker-compose.yml up -d
+# This script is primarily for CI purposes. If you want to run the e2e tests locally use the -L flag:
+#    ./e2e_test_setup.sh -L
+
+if getopts ":L" arg; then
+  mkdir -p nigiri
+  cd nigiri
+  if [[ ! ( -d "config" ) || ! ( -d "liquid-config" ) || ! ( -f "docker-compose.yml" ) ]]; then
+    curl https://travis.nigiri.network | bash
+  fi
```
bonomat, line 11: nice improvement :)

bonomat (approved):

PR #170, opened 2021-07-15
Add MacOS ci builds for wasm wallet tests.

bonomat:
We can add a MacOS ci build for the whole repository once we've removed the libraries (swap/covenants/etc). These modules use docker for their tests which is not available on MacOS. 

Note: LLVM 12 is already installed on MacOS (see [here](https://github.com/bonomat/droplet/runs/3074604513?check_suite_focus=true)), hence we only need to overwrite the environment variables.

CC: @rishflab / @thomaseizinger : this might be interesting for you.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,7 +129,15 @@ jobs:
 
 
   wasm_wallet_test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> Note: LLVM 12 is already installed on MacOS (see here), hence we only need to overwrite the environment variables.

Good finding, thanks for sharing!

da-kami (approved):

PR #186, opened 2021-07-22
Introduce extension page.

bonomat:
The extension page can be opened through clicking on the expand button in the popup. 
It will open in a new tab.


bonomat:
not relevant atm.

PR #189, opened 2021-07-27
Enable public listening over TLS

### LINKED ISSUE #190: [Waves] SSL certificate configuration

da-kami:
Fixes #190


bobtimus/src/bin/fake_bobtimus.rs
```diff
@@ -59,9 +61,24 @@ async fn main() -> Result<()> {
...
-                .await;
+            match (tls_certificate, tls_private_key) {
+                (Some(certificate), Some(private_key)) => {
+                    warp::serve(routes.or(faucet).with(cors))
+                        .tls()
+                        .cert_path(certificate)
+                        .key_path(private_key)
+                        .run(listen_address)
+                        .await;
+                }
```
bonomat, line 72: Does this serve http and https or is this now an or? 

We should serve http and https at the same time. E.g. if the provider wants to serve HTTPS and Tor. Tor needs to go against http only.

da-kami: Interestingly enough Tor + HTTPS [can actually be set up](https://community.torproject.org/onion-services/advanced/https/) - so I am tempted to say that, if a provider has a certificate he might as well just use that for the onion setup as well. 
I have, however, not tested this yet. Additionally I am not sure about the implications. Do you think it would make sense to validate that such a setup works (by running it) or would you just enforce serving it on both `http` and `https` if user configures it?
I think we can provide a separate listening address for https as well and enable serving both. As long as all components that keep state are fine with that (e.g. database being accessed from two different components) that should be fine. Are we using transactions to write into the DB already?

da-kami: After reading through the article I posted above one more time I think I take back what I said. We *will* have to serve on `http` and `https` - there are too many unknowns otherwise. I will see to change this accordingly.

bonomat: Just to clarify things: 

Assuming bobtimus has the domain `bob.comit.network` and has a certificate for the this domain he can run `https://bob.comit.network`. 
If he creates a Tor hidden service his service would be reachable under `thisisarandomononionaddress.onion`. 

Note that the certificate for `bob.comit.network` is not valid for `thisisarandomononionaddress.onion`. 

So if bobtimus would only accept https requests a request to **https**`://thisisarandomononionaddress.onion` will return an invalid certificate and a browser warning. 

In theory, bobtimus could get a certificate for `thisisarandomononionaddress.onion` but this is really hard (e.g. letsencrypt does not support this.) 

Hence, best practice is to disable https when provided hidden service which are reachable via an onion address.

da-kami: yeah, I also noticed that - he would need *two* certificates (with the second one being the one for the onion address...) and the is really not what is wanted. Already changed, I did the changes as `!fixup` in case you want to see them in isolation. Will squash that later :)

bobtimus/src/bin/bobtimus.rs
```diff
@@ -42,23 +41,19 @@ async fn main() -> Result<()> {
...
+                    .cert_path(https.tls_certificate)
+                    .key_path(https.tls_private_key)
+                    .run(https.listen_https)
+                    .await;
+            }
+
+            if let Some(listen_http) = http {
+                warp::serve(http::routes(bobtimus, subscription))
+                    .run(listen_http)
+                    .await;
```
bonomat, line 56: Did you try if this now listens on both? 
I think `warp::serve` is blocking and once `if let Some(https) == Some` the part below will never be executed.

da-kami:
```
run --package bobtimus --bin bobtimus -- start
--http 127.0.0.1:3030
--https 127.0.0.1:3033
--tls-private-key /Users/dakami/CoBloX/playground/SSL-cert/server.key
--tls-certificate /Users/dakami/CoBloX/playground/SSL-cert/server.crt
--elementsd http://admin1:123@127.0.0.1:7041
```


image

bobtimus/src/bin/bobtimus.rs
da-kami: I adapted the `bobtimus` code - did not bother with `fake_bobtimus` and added a `bail` for `https` instead.

bonomat:
Looks good. I wait with approval because mergify would just merge it and I think you wanted to squash those commits.

bonomat (approved):

PR #191, opened 2021-07-27
Create wallet from bip39 seed words

### LINKED ISSUE #188: [Extension] Wallet backup/recovery
- When creating a wallet the user should be able to receive a _thing_ which he can back up outside of our implementation
- The user should be able to create a wallet from that _thing_

_thing_ could be seed words or just the private key.

bonomat:
resolves #188 

- [x] Fix e2e tests (they are still using the old ui)

some screenshots: p.s. I apply again for COMIT design award 🤪

image

image

image


da-kami (commented):
There were two empty files committed 

```
 e2e_tests/jest.config.js
 e2e_tests/package.json
```

is that on purpose?

Are the derivation path TODOs something we have to action before using this on mainnet? In my understanding we don't have to, but when we updgrade the wallet old wallets might become unusable - correct?

bonomat:
> There were two empty files committed
> 
> ```
>  e2e_tests/jest.config.js
>  e2e_tests/package.json
> ```
> 
> is that on purpose?

The files are not empty but are also not changed... Looks like some sort of git error because I changed and unchanged things. 

> 
> Are the derivation path TODOs something we have to action before using this on mainnet? In my understanding we don't have to, but when we updgrade the wallet old wallets might become unusable - correct?

I don't think we have to do this now. The TODO means we should derive a new key for each address instead of only using the first derived one.

e2e_tests/src/e2e.test.ts
```diff
@@ -20,106 +19,33 @@ describe("webdriver", () => {
...
-        const service = new firefox.ServiceBuilder(firefoxPath);
-        const options = new firefox.Options().headless();
+        const debug = Debug("e2e-setup");
 
-        driver = new Builder()
-            .setFirefoxService(service)
-            .forBrowser("firefox")
-            .setFirefoxOptions(options)
-            .build();
+        let { driver: d, extensionId: eId, extensionTitle: eTitle, webAppTitle: wTitle } =
```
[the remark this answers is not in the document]
bonomat: Yea... I failed here. I don't know how to do this in a _nicer_ way ...

extension/src/components/CreateWallet.tsx
```diff
@@ -0,0 +1,156 @@
...
+const debug = Debug("unlock-wallet");
+
+type CreateWalletProps = {
+    onUnlock: () => void;
+};
+
+function CreateWallet({ onUnlock }: CreateWalletProps) {
+    const [backedUp, setBackedUp] = useState(false);
+    const [seedWords, setSeedWords] = useState(
+        "chase prevent symptom panel promote short tray cigar wonder vanish sustain hurry",
```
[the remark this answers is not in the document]
bonomat: This was for deterministic testing purposes. We can just remove the initial value.

da-kami (approved):
Don't see blockers concerning the implementation.

extension/wallet/Cargo.toml
```diff
@@ -15,6 +15,8 @@ aes-gcm-siv = { version = "0.9", features = [ "std" ] }
 anyhow = "1"
 baru = "0.1"
 bdk = { version = "0.4", default-features = false }
+bip32 = { version = "0.2", features = [ "secp256k1" ] }
+bip39 = { version = "1", features = [ "rand" ] }
```
thomaseizinger, line 19: The `bip32` crate comes with bip39 support, why an extra dependency?

extension/wallet/Cargo.toml
```diff
@@ -15,6 +15,8 @@ aes-gcm-siv = { version = "0.9", features = [ "std" ] }
 anyhow = "1"
 baru = "0.1"
 bdk = { version = "0.4", default-features = false }
+bip32 = { version = "0.2", features = [ "secp256k1" ] }
```
thomaseizinger, line 18: Chosing the `secp256k1` feature over the `secp256k1-ffi` feature pulls in another dependency in the form of `k256` whilest `secp256k1-ffi` would result in `rust-secp256k1` which we already have in the dependency graph.

thomaseizinger (commented):
Two things I noticed in regards to dependencies.

extension/wallet/Cargo.toml
bonomat: Addressed in follow-up PR here: https://github.com/comit-network/waves/pull/193

extension/wallet/Cargo.toml
bonomat: Addressed in follow-up PR here: https://github.com/comit-network/waves/pull/193

PR #194, opened 2021-07-28
Loan offer

### LINKED ISSUE #187: [Waves] Lending feature in UI
Current situation is that the lending feature in waves is not fully functional: 

- most values are hard-coded in the UI but should be originated from bobtimus
- there are calculation errors, e.g. if the loan collateral is not equal 1btc number format errors might show up (something about too many decimal places)

Goal: 
- A loan should be a concrete offer by bobtimus, e.g. bobtimus defines the rate/loan term/interest rate/collateral to principal ratio/min/max amount/etc.  (these values can be hardcoded in bobtimus for now)
- The lender can just accept it and change the principal amount.

da-kami:
Fixes #187 

We are reaching a mergeable state given what is descibed in the ticket.
Overall the lending feature is still a bit brittle. Id say we record tickets for specific "bugs" rather than keep extending this PR.


bobtimus/src/lib.rs
```diff
@@ -258,10 +260,34 @@ where
         Ok(transaction)
     }
 
+    /// Handle Alice's loan offer request
+    ///
+    /// We return the possible loan terms that we offer to Alice
+    /// Alice can then request a loan that is within the parameters that were offered
+    pub async fn handle_loan_offer_request(&mut self) -> Result<LoanOffer> {
```
bonomat, line 267: At some point we should remove the name "Alice" (and "Bob") from bobtimus and replace it with "borrower" (and "lender" respectively).

bobtimus/src/lib.rs
```diff
@@ -258,10 +260,34 @@ where
...
+        Ok(LoanOffer {
+            rate: self.rate_service.latest_rate(),
+            // TODO: Should be a proper fee estimation
+            fee_sats_per_vbyte: Amount::from_sat(1),
+            min_principal: LiquidUsdt::from_satodollar(10),
+            max_principal: LiquidUsdt::from_satodollar(10000),
+            min_ltv: dec!(1.5),
+            interest: vec![Interest {
+                // TODO: What is the blocktime on Liquid so we can map it to time interval?
+                timelock: 4320,
```
bonomat, line 277: Liquid has 1 minute blocks.

bobtimus/src/loan.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[serde(with = "::elements::bitcoin::util::amount::serde::as_sat")]
+    pub fee_sats_per_vbyte: Amount,
+
+    #[serde(serialize_with = "LiquidUsdt::serialize_to_nominal")]
+    pub min_principal: LiquidUsdt,
+    #[serde(serialize_with = "LiquidUsdt::serialize_to_nominal")]
+    pub max_principal: LiquidUsdt,
+
+    /// The minimum ltv that is acceptable in percent
+    pub min_ltv: Decimal, // TODO: is u64 appropriate here - how is ltv depicted internally?
```
bonomat, line 19: I thought I point this out: 

LTV => **L**oan **T**o **V**alue
Since we want to over-collateralize the loan the LTV ratio will be smaller than 0. 

e.g. 
- current rate: 1 BTC = $100
- principal amount = $100
- collateral amount = 2 BTC
--> LTV = 100 / 2*100 = 0.5

The higher this value is the higher the risk for the lender.

which means LTV value should be a floating point number and a max. 

Note: In the loan contract the liquidation is decided according to the exchange rate (https://github.com/comit-network/baru/blob/a7743b538c91ff3f055c3a1f22a14502be7e275b/src/loan.rs#L68) (cc @luckysori ). 
We will need to derive the value we need to put into liquidation function from the agreed max LTV ratio.

bobtimus/src/lib.rs
da-kami: Yeah, good to keep it in mind, I started doing it now (i.e. where I touched code that contains Alice/Bob I change it where needed) but I did not clean up all instances. Might be a bit tricky - in a protocol context we will want to leave it I suppose.

da-kami:
Should have run the tests first... those still need some fixing / extending.

extension/wallet/src/lib.rs
```diff
@@ -225,10 +225,22 @@ pub async fn make_sell_create_swap_payload(
 pub async fn make_loan_request(
     wallet_name: String,
     collateral: String,
+    fee_rate: String,
+    timeout: String,
 ) -> Result<JsValue, JsValue> {
-    let collateral = map_err_from_anyhow!(parse_to_bitcoin_amount(collateral))?;
+    // TODO: Change the UI to handle SATs not BTC
```
bonomat, line 240: I agree with this. Shall we record a ticket for this?

extension/wallet/src/lib.rs
```diff
@@ -225,10 +225,22 @@ pub async fn make_sell_create_swap_payload(
 pub async fn make_loan_request(
     wallet_name: String,
     collateral: String,
+    fee_rate: String,
+    timeout: String,
```
bonomat, line 238: What do you think of changing the API in a way that we always deal with days until we get back the loan response. Thatway Bobtimus has to the calculation to actual blockheight and we only need to confirm it again in the UI when we sign it?

extension/src/components/OpenLoans.tsx
```diff
@@ -1,15 +1,26 @@
...
 import * as React from "react";
 import { useAsync } from "react-async";
 import { repayLoan } from "../background-proxy";
+import { getBlockHeight } from "../background-proxy";
 import { LoanDetails } from "../models";
 import Btc from "./bitcoin.svg";
 import Usdt from "./tether.svg";
 
-const error = Debug("openloans:error");
+const debug = Debug("openloans:error");
```
bonomat, line 23: I prefer not renaming it. 
`debug` is for printing _normal_ messages
`error` is for printing _error_ messages.

bonomat (commented):
Cool, definitely well seen changes :)

bonomat (dismissed):
Only change request from me is on renaming the `error` to `debug` .

extension/src/components/OpenLoans.tsx
da-kami: I only changed this because the I got `yarn tsc` errors with the `error` variable. I'll give it another try :)

extension/wallet/src/lib.rs
da-kami: https://github.com/comit-network/waves/issues/205

extension/wallet/src/lib.rs
da-kami: Yes! I should have done that in the first place. The absolute timelock being decided at the very start is not optimal, because the relative time may already decline while the user is still deciding on the loan. This calculating the timelock upon taking that is not an issue anymore :)

da-kami: will do this in a follow-up that includes baru upgrades (because to do this properly some minor changes in baru will be needed that I'll just do on the way)

PR #195, opened 2021-07-29
Fake bobtimus is dead long live bobtimus

bonomat:
This simplifies our code-base as we now have only one binary to maintain. Also, the exchange rate in the UI are now real world values.


da-kami (commented):
I had a quick look and it looks good! 
I like the configuration as feature.
I will rebase my work on top of this.

PR #198, opened 2021-07-30
Deterministic e2e test setup

bonomat:
This PR aims at having deterministic e2e tests. For that we remove as much randomness from the tests and move it into the e2e_test_setup.sh. 

- each test has its own file and own setup. With this approach we don't run into the danger of reusing the same state in different tests which can make problems
- each test uses its own wallet. that way a test does not spend utxos which are used elsewhere
- tests are run sequentially. By having a separate file for each test we can enforce that the tests are run sequentially. This is slower but removes randomness.
- No faucets in tests: by moving the funding of the test wallets into the e2e test setup we remove one more source of randomness.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Run test
         run: |
           yarn install
-          yarn run test
+          yarn run all-tests
 
       - name: Archive logs
         uses: actions/upload-artifact@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,9 +60,7 @@ jobs:
           name: Logfiles
           path: |
             ./e2e_tests/bobtimus.log
-            ./e2e_tests/liquid.log
-            ./e2e_tests/electrs.log
-            ./e2e_tests/esplora.log
+            ./e2e_tests/screenshots
 
   build_test_workspace:
     strategy:
--- a/e2e_tests/e2e_test_setup.sh
+++ b/e2e_tests/e2e_test_setup.sh
@@ -12,7 +12,7 @@ while getopts "CS" o; do
             then
                 echo "Clearing docker compose data"
                 cd nigiri
-                docker compose down
+                docker-compose down
                 rm -rf liquid-config/liquidregtest
                 cd ..
             else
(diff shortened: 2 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -60,9 +60,7 @@ jobs:
           name: Logfiles
           path: |
             ./e2e_tests/bobtimus.log
-            ./e2e_tests/liquid.log
```
da-kami, line 63: Are those logs not used or not helpful; why remove them?

e2e_tests/src/utils.ts
```diff
@@ -62,32 +65,6 @@ const setupBrowserWithExtension = async (webAppUrl: string) => {
     };
 };
 
-// set testing wallet.
```
da-kami, line 58: bit confused why this was removed in this commit, but I reckon it was not needed anymore.

e2e_tests/e2e_test_setup.sh
```diff
@@ -75,3 +75,14 @@ echo "Bitcoin Asset ID: "$btc_asset_id
             --elementsd http://admin1:123@localhost:18884 \
             --usdt $usdt_asset_id > e2e_tests/bobtimus.log 2>&1 &
 )
+
+# give bobtimus time to start:
+
+sleep 5s
+
+# fund test wallets
+curl -POST  http://localhost:3030/api/faucet/el1qqfc6m8nrnrfpcsj9thg2df9jzpv4pwm4shm5wvn2untay2dmhhk4ut8j35jtd7jvqplx3gt9zth7ndtn6mcq68d03l4e7wqay -d {}
```
da-kami, line 84: it might help to add a comment here to know which test uses which wallet and which one is bobtimus (I assume the first one, because that address is not used in the wallet tests...)

da-kami (commented):
Wallet for each test is a bit cumbersome as you state, but if it makes the setup more stable that is good enough for now. Don't expect that there will be a big number of e2e tests on top of what we have, so the overhead should be fine.

.github/workflows/ci.yml
bonomat: These files simply do not exist :D

e2e_tests/src/utils.ts
bonomat: Sorry for that.

e2e_tests/e2e_test_setup.sh
bonomat: Well, I guess that's not needed anymore :D

PR #202, opened 2021-07-30
Extract stateless functions into wallet library module

rishflab:
Sorry for tagging everyone as reviewer but It kind of does concern everyone.

NOTE: probably easier to checkout the branch and look at it in your IDE

The goal of this commit/branch is to extract wallet
functionality into a module so it can moved into the baru repo The
motivation for this refactor is so the platform team can upgrade the
protocols to use PSET's. This should also take some workload away
 from from product team.

The stuff that will be moved into the baru repo is in the baru-wallet crate


extension/wallet/src/lib_wallet.rs
```diff
@@ -0,0 +1,348 @@
...
+            .into_iter()
+            .map(|(asset, utxos)| {
+                let total_sum = utxos.into_iter().sum();
+                BalanceEntry {
+                    asset,
+                    value: total_sum,
+                }
+            })
+            .collect()
+    }
```
thomaseizinger, line 87: Why is this not just a stand-alone function? Seems unnecessary to depend on `self` if you could also just take in the blinding-key?

extension/wallet/src/lib_wallet.rs
```diff
@@ -0,0 +1,348 @@
...
+
+                filter_map(utxo, txout)
+            })
+            .collect::<FuturesUnordered<_>>()
+            .filter_map(|r| std::future::ready(r.transpose()))
+            .try_collect::<Vec<_>>()
+            .await?;
+
+        Ok(txouts)
+    }
```
thomaseizinger, line 109: Same here, this should be standalone but with a type parameter `W: Wallet` so you can call `fetch_utxos` and `fetch_transaction`.

Alternatively, you could make this a function on an extension trait as well, the same way `FutureExt` adds functionality to everything that implements `Future`.

Default functions in traits only make sense if you want to potentially override them. There is no reason why we would override this.

extension/wallet/src/lib_wallet.rs
```diff
@@ -0,0 +1,348 @@
...
+    #[serde(with = "bdk::bitcoin::util::amount::serde::as_sat")]
+    pub amount: bdk::bitcoin::Amount,
+}
+
+// todo!(Make our own UTXO type, this is a http payload)
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize)]
+pub struct SwapUtxo {
+    pub outpoint: OutPoint,
+    pub blinding_key: SecretKey,
+}
```
thomaseizinger, line 296: This should be in the extension, not the wallet.

extension/wallet/src/lib_wallet/extract_loan.rs
```diff
@@ -0,0 +1,79 @@
...
+use baru::loan::{Borrower0, Borrower1, LoanResponse};
+use elements::{secp256k1_zkp::SECP256K1, AssetId};
+
+pub async fn extract_loan<W: Wallet + Sync>(
+    wallet: &W,
+    loan_response: LoanResponse,
+    btc_asset_id: AssetId,
+    usdt_asset_id: AssetId,
+    borrower0: Borrower0,
+) -> Result<(Borrower1, LoanDetails), Error> {
```
thomaseizinger, line 11: Let's try and make as many functions of this _data-driven_ instead of "dependency"-driven, i.e.:

Take in a list of utxos as a parameter and define your functionality based on that.

A good rule of thumb is: Only depend on functionality through a type parameter if it is conditional on something else. In this case, you can just pass in a list of UTXOs. A value-driven API is always more flexible which is what you want in a library.

extension/wallet/src/lib_wallet/get_balances.rs
```diff
@@ -0,0 +1,10 @@
+use crate::lib_wallet::{BalanceEntry, Wallet};
+use anyhow::Result;
+
+pub async fn get_balances<W: Wallet + Sync>(wallet: &W) -> Result<Vec<BalanceEntry>> {
+    let txouts = wallet.get_txouts(|_, txout| Ok(Some(txout))).await?;
+
+    let balances = wallet.compute_balances(&txouts);
+
+    Ok(balances)
+}
```
thomaseizinger, line 10: That seems to be a pretty useless function :D

extension/wallet/src/lib_wallet/get_transaction_history.rs
```diff
@@ -0,0 +1,13 @@
...
+use crate::lib_wallet::Wallet;
+
+pub async fn get_transaction_history<W: Wallet + Sync>(wallet: &W) -> Result<Vec<Txid>> {
+    // We have a single address, so looking for the transaction
+    // history of said address is sufficient
+    let address = wallet.get_address();
+    let history = wallet.fetch_transaction_history(&address).await?;
+
+    Ok(history)
+}
```
thomaseizinger, line 13: Same here.

thomaseizinger (commented):
Overall a good starting point. Try to push for more value-based APIs where possible instead of depending on traits.

Additionally, I think it should actually be the responsibility of `baru` to provide "backends" like esplora. So maybe the `Wallet` trait can be trimmed down to a `Blockchain` backend (similar to `bdk`) and then we define functionality composed on top of this trait?

Also, I think `baru` should offer an actual `Wallet` struct that the extension can re-use that holds state like the keys, and UTXOs. When given a `Backend`, the wallet can then `sync` the UTXOs and perform IO-less stuff like signing a loan transaction etc.

thomaseizinger (commented):
This is coming together nicely.

Take the following with a grain of salt because maybe it is not fully-possible but I was thinking we could do something like:

1. Have `Wallet` store a list of unspent transaction outputs. Whatever data structure we store, it should be everything we need to know about our UTXOs. In the future, we can abstract this into some kind of backend rather then a just an in-memory list.
2. Have a `Wallet::sync` function that looks roughly like this:

```rust
impl Wallet {
    async fn sync(&mut self, backend: &impl Backend) {
        let utxos = backend.fetch_utxos(self.address());

         // for utxo in utxos, fetch transaction and compute txout ...

         // self.txouts = 
    }
}
```

3. All APIs like `extract_loan` can now be synchronous functions on `Wallet` because `Wallet` owns most of the data necessary to do the task (txouts, blinding key etc)

Some notes about this design:

- It should allow the `Backend` trait to be minimal and only concern itself with IO.
- `baru` _can_ provide an implementation of `Backend` for esplora but given that specific need of caching using the browser's cache etc, I would probably not bother and leave that within the extension for now.
- Passing the `Backend` as a parameter to `sync` allows `Wallet` to remain simple and without type parameters and allows for easier testing using dummy backends.
- Having `Wallet` own state like `TxOuts` makes all the functions like `extract_loan` quite easy to use.

On top of all this, the extension should probably define its own `Wallet` struct that wraps the one provided by baru and carry around the state for usdt_asset_id, chain, etc.

---

I recognize this may seem contradicting to my previous review. However, for this design to be possible, it was important to first split things into pure functions (value-based). Now that this is the case, we can easily re-compose everything that is there by bundling the necessary state for these functions up in the `Wallet` struct.

rishflab:
> * Having `Wallet` own state like `TxOuts` makes all the functions like `extract_loan` quite easy to use.

good idea. another benefit is the caller won't have to pass around the touts in their app

baru_wallet/src/client.rs
```diff
@@ -0,0 +1,9 @@
+use crate::Utxo;
+use anyhow::Result;
+use async_trait::async_trait;
+use elements::{Address, TxOut};
+
+#[async_trait]
+pub trait Client {
```
thomaseizinger, line 7: We could follow the "single trait fn" pattern here and name this `GetTxouts`.

Plus, I think the `Utxo` type should be moved into this file.

baru_wallet/src/lib.rs
```diff
@@ -0,0 +1,652 @@
...
+#[derive(Debug, thiserror::Error)]
+#[error("Unsupported chain: {0}")]
+pub struct WrongChain(String);
+
+#[derive(Debug)]
+pub struct Wallet {
+    pub name: String,
+    encryption_key: [u8; 32],
+    secret_key: SecretKey,
+    pub sk_salt: [u8; 32],
```
thomaseizinger, line 76: Do these have to be public?

baru_wallet/src/lib.rs
```diff
@@ -0,0 +1,652 @@
...
+
+impl Wallet {
+    pub async fn sync(&mut self, client: &impl Client) -> Result<()> {
+        self.cache = client.get_txouts(self.address()).await?;
+        Ok(())
+    }
+    pub fn utxos(&self) -> Vec<(Utxo, TxOut)> {
+        self.cache.clone()
+    }
+    pub fn initialize_new(
```
thomaseizinger, line 91: Missing newline between a couple of these functions.

baru_wallet/src/lib.rs
```diff
@@ -0,0 +1,652 @@
...
+const SECRET_KEY_ENCRYPTION_NONCE: &[u8; 12] = b"SECRET_KEY!!";
+
+impl Wallet {
+    pub async fn sync(&mut self, client: &impl Client) -> Result<()> {
+        self.cache = client.get_txouts(self.address()).await?;
+        Ok(())
+    }
+    pub fn utxos(&self) -> Vec<(Utxo, TxOut)> {
+        self.cache.clone()
+    }
```
thomaseizinger, line 90: I don't think we should expose our UTXOs. That reveals too much of the inner structure of this type.

baru_wallet/src/lib.rs
```diff
@@ -0,0 +1,652 @@
...
+
+        Self {
+            collateral,
+            principal_repayment: principal.amount,
+            principal,
+            term: timelock,
+            txid,
+        }
+    }
+}
```
thomaseizinger, line 530: Does this really have to live in the `baru`?

extension/wallet/src/protocol/extract_loan.rs
```diff
@@ -1,16 +1,35 @@
...
+/// "handshake" so that we can later on sign the loan transaction and
+/// give it back to the lender.
+#[wasm_bindgen]
+pub async fn extract_loan(wallet_name: String, loan_response: JsValue) -> Result<JsValue, JsValue> {
+    let loan_response = map_err_from_anyhow!(loan_response.into_serde())?;
+    let loan_details = map_err_from_anyhow!(extract_loan_w(wallet_name, loan_response).await)?;
+    let details = map_err_from_anyhow!(JsValue::from_serde(&loan_details))?;
+
+    Ok(details)
+}
```
thomaseizinger, line 29: Does this work if the `wasm_bindgen` annotated function is not at the root of the library?

extension/wallet/src/protocol/extract_loan.rs
```diff
@@ -72,23 +83,27 @@ pub async fn extract_loan(
...
         collateral_balance,
         usdt_asset_id,
-        borrower.principal_tx_out_amount,
+        borrower1.principal_tx_out_amount,
         principal_balance,
         timelock,
         loan_txid,
-    )
-    .map_err(Error::LoanDetails)?;
+    );
```
thomaseizinger, line 101: Looks to me like `LoanDetails` should live in here.

extension/wallet/src/protocol/extract_loan.rs
```diff
@@ -1,16 +1,35 @@
...
+#[wasm_bindgen]
+pub async fn extract_loan(wallet_name: String, loan_response: JsValue) -> Result<JsValue, JsValue> {
+    let loan_response = map_err_from_anyhow!(loan_response.into_serde())?;
+    let loan_details = map_err_from_anyhow!(extract_loan_w(wallet_name, loan_response).await)?;
+    let details = map_err_from_anyhow!(JsValue::from_serde(&loan_details))?;
+
+    Ok(details)
+}
+
+pub async fn extract_loan_w(
```
thomaseizinger, line 31: Unnecessarily public?

extension/wallet/src/protocol/make_create_swap_payload.rs
```diff
@@ -23,54 +36,76 @@ pub async fn make_buy_create_swap_payload(
...
+        EsploraClient::new(guard.clone())
+    };
+
+    let (utxos, secret_key, address, blinding_key) = {
+        let mut wallet = current(&wallet_name, &LOADED_WALLET)
+            .await
+            .map_err(Error::LoadWallet)?;
+        wallet.sync(&client).await.map_err(Error::SyncWallet)?;
+        (
+            wallet.utxos(),
```
thomaseizinger, line 87: Hmm, what is the point of calling `sync` if the only thing we do is take the UTXOs out again? :sweat_smile: 

If you are unsure on how much functionality should move into the wallet here, then just do this:

```rust
let utxos = client.get_utxos(wallet.address()).await?;
```

Although ideally, we can add a better API to `Wallet` that reduces the amount of code we have to write in here.

What we are doing in here is coin-selection. Can we have a `Wallet::coin_select` API that returns a list of UTXOs that is enough to fund a certain amount?

extension/wallet/src/protocol/repay_loan.rs
```diff
@@ -137,13 +152,8 @@ pub async fn repay_loan(
...
                 // protocols
-                let script_witness = sign_with_key(
-                    SECP256K1,
-                    &mut cache,
-                    index,
-                    &wallet.secret_key,
-                    output.value,
-                );
+                let script_witness =
+                    sign_with_key(SECP256K1, &mut cache, index, &secret_key, output.value);
```
thomaseizinger, line 156: Can we add some form of signing API to `Wallet` that reduces the amount of code we have to write here?

thomaseizinger (changes requested):
Looking good, I think we are almost there!

To summarize:

- Add a coin-selection API to `Wallet`
- Add a signing API to `Wallet`

That should allow you to remove the `utxos()` accessor on `Wallet` and greatly reduce the duplication between the exposed wallet functions.

extension/wallet/src/protocol/extract_loan.rs
rishflab: not sure, was hoping someone else would know the answer to this

extension/wallet/src/protocol/extract_loan.rs
rishflab: used all over the place

extension/wallet/src/protocol/make_create_swap_payload.rs
rishflab: yep agreed, I think coin select should be pretty easy to move to the wallet

extension/wallet/src/protocol/extract_loan.rs
thomaseizinger: Well, if the tests still run, we should be good :)

extension/wallet/Cargo.toml
```diff
@@ -13,13 +13,13 @@ default = ["console_error_panic_hook"]
 [dependencies]
 aes-gcm-siv = { version = "0.9", features = ["std"] }
 anyhow = "1"
-baru = { git = "https://github.com/da-kami/baru", branch = "lender-decides-timelock" }
 bip32 = { version = "0.2", features = ["secp256k1-ffi", "bip39"], default-features = false }
-coin_selection = { path = "../../coin_selection" }
+async-trait = "0.1"
+baru = { path = "/Users/rishflab/comit-network/baru"}
+bdk = { version = "0.4", default-features = false }
```
thomaseizinger, line 19: This is a really old version of bdk. The current version is 0.8: https://github.com/bitcoindevkit/bdk/releases

extension/wallet/src/lib.rs
```diff
@@ -84,65 +168,6 @@ pub fn setup() {
...
-
-/// Retrieve the status of the wallet with the given name.
-#[wasm_bindgen]
-pub async fn wallet_status(name: String) -> Result<JsValue, JsValue> {
-    let status = map_err_from_anyhow!(wallet::get_status(name, &LOADED_WALLET).await)?;
-    let status = map_err_from_anyhow!(JsValue::from_serde(&status))?;
-
-    Ok(status)
-}
-
```
thomaseizinger, line 145: I know that it is a result of our iterations but it would be nice to keep the diff small where possible (also to make your own life easier in resolving merge conflicts).

From what I can see, this part of the code didn't actually functionally change, it was just moved into dedicated files.

extension/wallet/src/protocol/make_create_swap_payload.rs
```diff
@@ -0,0 +1,158 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Wallet is not loaded: {0}")]
+    LoadWallet(anyhow::Error),
+    #[error("Could not sync waller: {0}")]
+    SyncWallet(anyhow::Error),
+    #[error("Coin selection: {0}")]
+    CoinSelection(anyhow::Error),
+}
```
thomaseizinger, line 131: Worth looking into the configuration options of `thiserror`: https://docs.rs/thiserror/1.0.26/thiserror/index.html#details

Instead of printing the error inline using `{0}`, you would be better of annotating it with `#[source]` to print it as part of the cause chain.

extension/wallet/Cargo.toml
luckysori: Pretty sure we don't need this `bdk` dependency here. But we can look to bump this version in `baru`'s `coin_selection` module.

extension/wallet/src/protocol/extract_trade.rs
```diff
@@ -1,38 +1,54 @@
...
 
-    let blinding_key = wallet.blinding_key();
+        let utxos = wallet.utxos();
+        let balances = wallet.compute_balances();
+        let blinding_key = wallet.blinding_secret_key();
+        let address = wallet.address();
+        (utxos, balances, blinding_key, address)
+    };
 
     let our_inputs = transaction
```
thomaseizinger, line 47: Finding our inputs (and our outputs) in a transaction could be considered generic enough wallet functionality so we can include it in the wallet lib.

My goal would be to get rid of `Wallet::utxos`. I think it would be really good for encapsulation (and hence API evolvement) if we would not expose that.

thomaseizinger: For example, if we ever want to change our wallet to not reuse addresses or change how blinding keys are related to addresses, this code would have to be updated in `waves`. But if the wallet is in control of the task of "identify my inputs and outputs", we can make the change transparently.

extension/wallet/src/protocol/repay_loan.rs
```diff
@@ -0,0 +1,150 @@
...
+        let mut wallet = current(&wallet_name, &LOADED_WALLET)
+            .await
+            .map_err(Error::LoadWallet)?;
+        wallet
+            .sync(&client.clone())
+            .await
+            .map_err(Error::SyncWallet)?;
+        Ok(wallet.sign(transaction))
+    };
+
```
thomaseizinger, line 80: `Wallet::sign` and `Wallet::coin_selection` both take `&self`. I think you should be able to get away with:

1. Unlocking the wallet only once.
2. Syncing the wallet only once.
3. Passing simple, _non_-move async blocks into `loan_repayment_transaction` for coin-selection and signer. Because the references to wallet are not mutable, you can have two at the same time.

thomaseizinger (commented):
The diff here is quite big :/

What is your appetite for making it smaller?
I think there might be a reasonably easy way of doing that.

1. Make a new branch from master.
1. First commit: Bump the library to `baru` version `0.3`.
1. Second commit: Introduce `EsploraClient` as a struct that owns the `base_url`
1. Third commit: Add a git dependency on `baru` that uses the new wallet and deletes all the old stuff.

Given that the overall structure of the `#[wasm_bindgen]` functions etc didn't change much, (4) should be relatively straight forward and should result in a much smaller diff :)

extension/wallet/src/protocol/make_create_swap_payload.rs
rishflab: leaving this to the platform team

thomaseizinger (approved):
Nice work! This diff is actually reviewable now :)

da-kami:
> You probably want to rebase your stuff onto #233 and drop the commit where you upgrade to `baru 0.3`.

I think the only commit that can be dropped is https://github.com/comit-network/waves/pull/202/commits/95fade9cf528c5d553a80171fcbd5267db5d4ae5 (the one that upgrades bobtimus)

The wallet upgrade should please stay - I don't touch that in my PR, but kept the deprecated API.

extension/wallet/Cargo.toml
```diff
@@ -13,7 +13,8 @@ default = ["console_error_panic_hook"]
 [dependencies]
 aes-gcm-siv = { version = "0.9", features = ["std"] }
 anyhow = "1"
-baru = "0.3"
+async-trait = "0.1"
+baru = { git = "https://github.com/comit-network/baru.git", branch = "extract-wallet" }
```
thomaseizinger, line 17: The branch has been merged right? So we can depend on latest HEAD now :)

PR #226, opened 2021-08-05
Create a loan backup file when confirming

bonomat:
Add loan back up feature to extension. 

image

image

image

image


da-kami:
seems this needs a proper rebase on master - looks like you built on top of Thomas' work but that was force pushed, or are Thomas' commits supposed to be here?

da-kami (commented):
Looks good - might want to tell the user that the backup is downloaded as file in the browser downloads. Not sure how feasible that is though.

One thing I find somewhat problematic: The stepper does not keep the state. If the extension is collapsed (also happens when changing focus away from the browser window to check the backup on file system...) - I have to press sign again (i.e.go through all the steps... - this is a bit weird, because in the end I can actually *not* check the file because that leads to an endless loop of doing all the steps....

extension/wallet/src/wallet/loan_backup.rs
```diff
@@ -0,0 +1,67 @@
...
+    name: String,
+    current_wallet: &Mutex<Option<Wallet>>,
+    txid: Txid,
+) -> Result<BackupDetails, Error> {
+    let storage = Storage::local_storage().map_err(Error::Storage)?;
+
+    // We get a hold of the wallet to ensure that it is loaded. This is a security mechanism
+    // to ensure no unauthorized access to the data.
+    // Ideally all data is encrypted but that's just how it is :)
+    let _ = current(&name, current_wallet).await.unwrap();
```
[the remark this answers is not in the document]
bonomat: I guess so. I created a ticket for that: https://github.com/comit-network/waves/issues/234

bonomat:
> One thing I find somewhat problematic: The stepper does not keep the state. If the extension is collapsed (also happens when changing focus away from the browser window to check the backup on file system...) - I have to press sign again (i.e. go through all the steps...) - this is a bit weird, because in the end I can actually not check the file because that leads to an endless loop of doing all the steps...

I know. @thomaseizinger, do you have an idea here? 
The issue is that the loan signing process is a 3 step process. If the user interupts it by clicking outside of the window, it starts from scratch. 
Shall keep some state here? 
A "simple" solution would be to open a popup window. That one will not close if you lose focus.

extension/src/background/index.ts
```diff
@@ -171,11 +173,32 @@ window.signLoan = async () => {
     // on the pop-up matches what is stored in the extension's
     // storage. It would be better to send around the swap ID to check
     // that the wallet is signing the same transaction the user has authorised
-    signLoan(walletName)
-        .then(resolveLoanSignRequest)
-        .catch(rejectLoanSignRequest)
-        .then(cleanupPendingLoan);
+
+    // if we receive an error, we respond directly, else we return the details
+    return await signLoan(walletName).catch(rejectLoanSignRequest);
```
thomaseizinger, line 178: `return await` is unnecessary, you can just remove the `await` and return the promise directly.

extension/src/components/ConfirmLoanWizard.tsx
```diff
@@ -19,99 +38,148 @@ const ConfirmLoanWizard = ({ onCancel, onSuccess, loanToSign }: ConfirmLoanWizar
...
             <Steps activeStep={activeStep}>
                 <Step label={"Sign"} key={"sign"} icon={FiClipboard}>
                     <Flex py={4}>
-                        <ConfirmLoan
-                            loanToSign={loanToSign}
-                            onCancel={onCancel}
-                            onSigned={onSigned}
-                        />
+                        <form
+                            onSubmit={async e => {
```
thomaseizinger, line 73: Doesn't need to be async. I think there is a typescript lint for such cases!

bonomat:
> > One thing I find somewhat problematic: The stepper does not keep the state. If the extension is collapsed (also happens when changing focus away from the browser window to check the backup on file system...) - I have to press sign again (i.e. go through all the steps...) - this is a bit weird, because in the end I can actually not check the file because that leads to an endless loop of doing all the steps...
> 
> I know. @thomaseizinger, do you have an idea here?
> The issue is that the loan signing process is a 3 step process. If the user interupts it by clicking outside of the window, it starts from scratch.
> Shall keep some state here?
> A "simple" solution would be to open a popup window. That one will not close if you lose focus.

I created a ticket for this: https://github.com/comit-network/waves/issues/235

extension/src/components/ConfirmLoanWizard.tsx
```diff
@@ -51,9 +52,7 @@ const ConfirmLoanWizard = ({ onCancel, onSuccess, loanToSign }: ConfirmLoanWizar
             const loanBackup = await createLoanBackup(signedTransaction);
             const file = new Blob([JSON.stringify(loanBackup)], { type: "text/json" });
             const url = URL.createObjectURL(file);
-            // Note: it would be nicer to open a dialog to download the file. However, we would lose focus
-            // of the window. We don't want that, hence we just open a new tab with the content
-            window.open(url, "_blank");
```
thomaseizinger: Nice!

This commit is candidate for squashing into the earlier one!

thomaseizinger: I can recommend looking into https://github.com/tummychow/git-absorb for a semi-automated solution to this.

thomaseizinger (commented):
> I know. @thomaseizinger, do you have an idea here?
> The issue is that the loan signing process is a 3 step process. If the user interupts it by clicking outside of the window, it starts from scratch.
> Shall keep some state here?

What about the following:

Instead of having backup be a dedicated action, we could add a checkbox like:

- [ ] Create backup of loan

When ticked, the download will automatically be initiated as part of clicking `sign`.

thomaseizinger:
Whoops, looks like my review is coming a bit too late, sorry about that, I only got a notification this morning!

In any case, I didn't see anything major that would have need to be addressed.

On a stylistic note: The guide we are striving to follow for commit messages [avoids dots at the end](https://chris.beams.io/posts/git-commit/#end) to a) save space and b) be consistent with git's internal commit messages like merges.

bonomat:
> Whoops, looks like my review is coming a bit too late, sorry about that, I only got a notification this morning!
> 
> In any case, I didn't see anything major that would have need to be addressed.
> 
> On a stylistic note: The guide we are striving to follow for commit messages [avoids dots at the end](https://chris.beams.io/posts/git-commit/#end) to a) save space and b) be consistent with git's internal commit messages like merges.

Sorry for merging so fast. was just in a call with Daniel and we decided to merge. I'll have a look at your feedback and get this sorted in follow-up PR(s).

extension/wallet/src/lib.rs
```diff
@@ -265,6 +265,35 @@ pub async fn sign_loan(wallet_name: String) -> Result<JsValue, JsValue> {
...
+}
+
+/// Load a loan backup and store it in localStorage.
+#[wasm_bindgen]
+pub async fn load_loan_backup(backup_details: JsValue) -> Result<(), JsValue> {
+    let backup_details: BackupDetails = map_err_from_anyhow!(backup_details.into_serde())?;
+    map_err_from_anyhow!(wallet::load_loan_backup(backup_details))?;
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
bonomat: I wonder if it is a good idea to call things directly from TS code instead. 
Are you not afraid that the TS code and rust code diverge too much? 

For example, the loan data is currently stored in localstorage under the key `loan_state`. If this ever to changes in the rust code we might run in the danger that it's not changed in the TS code.

thomaseizinger:
> > Whoops, looks like my review is coming a bit too late, sorry about that, I only got a notification this morning!
> > In any case, I didn't see anything major that would have need to be addressed.
> > On a stylistic note: The guide we are striving to follow for commit messages [avoids dots at the end](https://chris.beams.io/posts/git-commit/#end) to a) save space and b) be consistent with git's internal commit messages like merges.
> 
> Sorry for merging so fast. was just in a call with Daniel and we decided to merge. I'll have a look at your feedback and get this sorted in follow-up PR(s).

No sweat, take it as you wish :)

extension/wallet/src/lib.rs
```diff
@@ -265,6 +265,35 @@ pub async fn sign_loan(wallet_name: String) -> Result<JsValue, JsValue> {
...
+/// Create a json object which can be used to backup the loan details to the transaction
+///
+/// TODO: Pass in transaction id instead of transaction
+/// For now, we pass in the whole transaction just to access it's transaction id.
+/// This is because `sign_loan` returns the whole transaction and we have no means to deserialize
+/// the transaction from within TS.
+#[wasm_bindgen]
+pub async fn create_loan_backup(
+    wallet_name: String,
+    transaction: JsValue,
```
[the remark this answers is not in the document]
bonomat: Good question. It's a mix right now: 

```
{
 inner: "02000...."
}
```

As far as I remember we had challenges to serialize this to a simple string.

extension/src/components/ConfirmLoanWizard.tsx
bonomat: There are quite a lint errors. I created a ticket for that: https://github.com/comit-network/waves/issues/236

extension/wallet/src/lib.rs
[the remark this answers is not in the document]
bonomat: I looked into it: in this particular case it's unfortunately a bigger task : 

The issue here is that we need to know the transaction ID to load this information from the localstorage. All we have available is the whole transaction. We'd need to change quite a few things for this. It's worth it and should be done but since it's working atm I recorded an issue for this. #238

PR #227, opened 2021-08-06
Improve error handling between extension scripts

thomaseizinger:
The big picture of this change is to treat content-script and background-script as "dumb" pipes that basically just pass through data.

This allows an error that is returned from the wallet to pop-out in the in-page script without further modification.


extension/src/background/index.ts
```diff
@@ -26,92 +26,85 @@ import {
...
+addRpcMessageListener(
+    "makeLoanRequestPayload",
+    ([collateral, feerate]) => makeLoanRequestPayload(walletName, collateral, feerate),
+);
+
+addRpcMessageListener("signAndSendSwap", ([txHex]) => {
+    extractTrade(walletName, txHex).then(decoded => {
+        swapToSign = { txHex, decoded };
+        updateBadge();
+    });
```
thomaseizinger, line 76: Just realized that any error here will not be forwarded to the content script, need to fix!

extension/src/in-page/index.ts
```diff
@@ -1,223 +1,108 @@
...
-                }
-            };
-            window.addEventListener("message", listener);
-        });
-        window.postMessage({
-            kind: MessageKind.WalletStatusRequest,
-            direction: Direction.ToBackground,
-        }, "*");
-        return promise;
+        return invokeContentScript("walletStatus", []);
```
thomaseizinger, line 12: This can be implemented using a JS proxy in the future if it gets too cumbersome. Can easily happen in a follow-up!

extension/src/background/index.ts
```diff
@@ -135,57 +128,60 @@ window.getSwapToSign = async () => {
...
-        payload = await signLoan(walletName);
-    } catch (e) {
-        error(e);
-        err = e;
-    }
-
-    browser.tabs.sendMessage(tabId, { direction: Direction.ToPage, kind: MessageKind.SignedLoan, payload, error: err });
-    loanToSign = undefined;
-    updateBadge();
+    signLoan(walletName).then(resolveLoanSignRequest).catch(rejectLoanSignRequest).then(() => {
```
bonomat, line 169: If `catch` is called, we do not call `updateBadge` and the extension will always show the badge sign.

bonomat (commented):
Awesome stuff. Definitely easier now. 
I think we should fix the badge issue. This can be quite annoying.

extension/src/background/index.ts
thomaseizinger: Ah yes good _catch_! Will fix.

This state management is something I am hoping to fix by introducing a popup for signing stuff. However, as noted in https://github.com/comit-network/waves/issues/228, we can't use the wallet from two different pages at the moment because it would be loaded again due to how the `wasm-proxy` works at the moment.

extension/src/background/index.ts
```diff
@@ -26,92 +26,85 @@ import {
...
+
+addRpcMessageListener("signAndSendSwap", ([txHex]) => {
+    extractTrade(walletName, txHex).then(decoded => {
+        swapToSign = { txHex, decoded };
+        updateBadge();
+    });
+
+    return new Promise((resolve, reject) => {
+        resolveSwapSignRequest = resolve;
+        rejectSwapSignRequest = reject;
```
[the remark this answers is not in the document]
thomaseizinger: Where does this error surface? Can we check on a function whether or not it is still callable? I think we may just have to try anyway?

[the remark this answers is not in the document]
bonomat: I just tried it. It actually fails on the other side.

Steps: 
1. click take loan and wait until badge updates
2. refresh webpage (error below appears)
3. open popup: still pending loan. If I click `cancel`/`sign` the request disapears and nothing happens.  

```
Promise rejected after context unloaded: Actor 'Conduits' destroyed before query 'RuntimeMessage' was resolved
contentScript.bundle.js:4096
    invokeBackgroundScriptRpc moz-extension://6cad53cc-191c-8e46-982e-49664f2bb67b/contentScript.bundle.js:4096
```

So I guess it's all good?

extension/src/background/index.ts
thomaseizinger: Should be fixed in https://github.com/comit-network/waves/pull/227/commits/318d8c4646c79243e2e8cbc43cd73d1d7986be14.

extension/src/background/index.ts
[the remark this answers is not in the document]
thomaseizinger: I don't think there is much we can do if the web page is reloaded. By using single-use listeners with randomly generated request IDs, we are depending on the state of the webpage to wait for our response.

In a way, one could argue that this is cleaner: When I reload a web page, I want it to throw away all its state and start again. It is like restarting a computer.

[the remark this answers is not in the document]
bonomat: Agreed :)

bonomat (approved):
Looks good. Thanks 👍

bonomat:
Tested e2e tests manually ✔

PR #233, opened 2021-08-11
Upgrade Botimus to handle loan calc and validation

da-kami:
Lender now includes the loan calculations and the validation of the loan.
I clicked it through and it looks sane :)

To be done in follow ups:

- Send **all** amounts in sats and do conversion on UI side (did not tackle all amounts yet)
- Validation step for the loan term
- Upgrade the UI so the borrower can select values

### DIFF
--- a/bobtimus/src/amounts.rs
+++ b/bobtimus/src/amounts.rs
@@ -5,7 +5,7 @@ use rust_decimal::{
     Decimal, RoundingStrategy,
 };
 use serde::{Deserialize, Serialize};
-use std::{convert::TryFrom, fmt::Debug};
+use std::{convert::TryFrom, fmt, fmt::Debug};
 
 /// Prices at which 1 L-BTC will be traded, in L-USDt.
 ///
@@ -62,7 +62,7 @@ impl Rate {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Serialize, Default)]
+#[derive(Clone, Copy, PartialEq, PartialOrd, Eq, Serialize, Deserialize, Default)]
 pub struct LiquidUsdt(#[serde(with = "::elements::bitcoin::util::amount::serde::as_sat")] Amount);
 
 impl LiquidUsdt {
@@ -125,8 +125,17 @@ impl TryFrom<f64> for LiquidUsdt {
     }
 }
 
+impl fmt::Display for LiquidUsdt {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.0.fmt_value_in(f, Denomination::Bitcoin)?;
+        write!(f, " L-USDT")
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
-pub struct LiquidBtc(#[serde(with = "::elements::bitcoin::util::amount::serde::as_sat")] Amount);
+pub struct LiquidBtc(
+    #[serde(with = "::elements::bitcoin::util::amount::serde::as_sat")] pub Amount,
+);
 
 impl From<Amount> for LiquidBtc {
     fn from(amount: Amount) -> Self {
--- a/bobtimus/src/lib.rs
+++ b/bobtimus/src/lib.rs
@@ -43,9 +43,14 @@ pub mod loan;
 pub mod problem;
 pub mod schema;
 
-use crate::loan::{Interest, LoanOffer, LoanRequest};
+use crate::loan::{
+    calculate_interest_rate, calculate_liquidation_price, calculate_ltv,
+    calculate_repayment_amount, calculate_request_price, validate_loan_is_acceptable,
+    Collateralization, LoanOffer, LoanRequest, Term,
+};
 pub use amounts::*;
 use elements::bitcoin::PublicKey;
+use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use std::{
     convert::TryFrom,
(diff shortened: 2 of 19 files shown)


bobtimus/src/lib.rs
```diff
@@ -270,32 +275,92 @@ where
     /// We return the range of possible loan terms to the borrower.
     /// The borrower can then request a loan using parameters that are within our terms.
     pub async fn handle_loan_offer_request(&mut self) -> Result<LoanOffer> {
```
bonomat, line 277: Does not look like it needs to be async anymore.

thomaseizinger: There is a lint you can enable to catch these automatically: https://rust-lang.github.io/rust-clippy/master/index.html#unused_async

bobtimus/src/amounts.rs
```diff
@@ -5,7 +5,7 @@ use rust_decimal::{
     Decimal, RoundingStrategy,
 };
 use serde::{Deserialize, Serialize};
-use std::{convert::TryFrom, fmt::Debug};
+use std::{convert::TryFrom, fmt, fmt::Debug};
```
thomaseizinger, line 8: I tend to always import just the `fmt` module and refer to everything else from there. Makes it more consistent and reduces the number of imports.

The thing is, you will always need `fmt::Fromatter` and `fmt::Result` as well.

bobtimus/src/lib.rs
```diff
@@ -270,32 +275,92 @@ where
...
+                    collateralization: dec!(2.0),
+                    interest_mod: Decimal::ZERO,
+                },
+            ],
+        }
     }
 
     /// Handle the borrower's loan request in which she puts up L-BTC as
     /// collateral and we lend L-USDt to her which she will have to
     /// repay in the future.
```
thomaseizinger, line 324: Similarly to commit messages, idiomatic rustdoc has a subject line with more fine-grained documentation in another paragraph: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

bobtimus/src/lib.rs
```diff
@@ -270,32 +275,92 @@ where
...
+    pub async fn handle_loan_request(&mut self, loan_request: LoanRequest) -> Result<LoanResponse> {
+        let loan_offer = self.current_loan_offer();
+
+        let interest_rate = calculate_interest_rate(
+            loan_request.term,
+            loan_request.collateralization,
+            &loan_offer.terms,
+            &loan_offer.collateralizations,
+            loan_offer.base_interest_rate,
+        )?;
```
thomaseizinger, line 334: These might be worth extending with `.context` so you know which part failed.

extension/src/components/ConfirmLoan.tsx
```diff
@@ -28,20 +26,13 @@ export default function ConfirmLoan(
...
-        promiseFn: getBlockHeight,
-        onReject: (e) => debug("Failed to fetch block height %s", e),
-    });
-    let { data: blockHeight, reload: reloadBlockHeight } = blockHeightHook;
-
+    let [timestamp, setTimestamp] = useState(Math.floor(Date.now() / 1000));
     useInterval(() => {
-        reloadBlockHeight();
+        setTimestamp(Math.floor(Date.now() / 1000));
     }, 6000); // 1 min
```
thomaseizinger, line 32: Could extract this into a `useTime` hook given that you are using it two times.

bobtimus/src/loan.rs
```diff
@@ -44,42 +45,520 @@ pub struct LoanOffer {
...
+    }
+
+    let mut collateralization_interest_mod = Decimal::ZERO;
+    for collateralization in collateralization_thresholds {
+        if borrower_collateralization >= collateralization.collateralization {
+            collateralization_interest_mod = collateralization.interest_mod;
+            continue;
+        }
+        break;
+    }
```
bonomat, line 109: I don't understand the logic here :) Can you explain this please?

bobtimus/src/loan.rs
```diff
@@ -44,42 +45,520 @@ pub struct LoanOffer {
...
+    );
+
+    Ok(repayment_amount)
+}
+
+pub fn calculate_liquidation_price(
+    repayment_amount: LiquidUsdt,
+    collateral_amount: LiquidBtc,
+) -> Result<LiquidUsdt> {
+    let repayment_amount = Decimal::from(repayment_amount.as_satodollar());
```
bonomat, line 146: Even though the calculation look relatively simple it would be worth it to add a comment mentioning the whole equation.

bobtimus/src/loan.rs
```diff
@@ -44,42 +45,520 @@ pub struct LoanOffer {
...
-    /// Loan term in days
-    pub term: u32,
-    /// Collateralization in percent
-    ///
-    /// Rational: If a borrower over-collateralizes with e.g. 150% -> better rate than at 140%
+#[derive(Debug, Clone, Copy, serde::Serialize)]
+pub struct Term {
+    pub days: u32,
+    /// Interest to be added on top of the base interest rate for this term
+    pub interest_mod: Decimal,
```
bonomat, line 62: What is this needed for?

bobtimus/src/loan.rs
```diff
@@ -44,42 +45,520 @@ pub struct LoanOffer {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::proptest;
+    use rust_decimal::prelude::FromPrimitive;
+    use rust_decimal_macros::dec;
+
+    #[test]
+    fn test_calculate_interest_rate() {
```
bonomat, line 309: I honestly cannot follow your tests. 
Given that these calculations are critical can we have some code comments explaining what is happening?

bobtimus/src/loan.rs
da-kami: Overall idea: The lender can define different interest rates based on term and collateralization. The `LoanOffer` contains term and collateralization thresholds that may modify the interest rate of the loan. For collateralizations the assumption is that the modifier would usually be `-x` (if higher collateraliztion then lower interest rate), for the term it would be `+x` (if longer term then higher interest rate).  

This piece of code: The modifier is set inside the if. As long as the collateralization falls into a higher threshold we keep looping. If the threshold is not higher we stop and use the modifier that was last set.

Note that for the term this has to be more restrictive - I will add a check to the loan acceptance criteria for that.

bobtimus/src/loan.rs
da-kami: see https://github.com/comit-network/waves/pull/233#discussion_r687388697

bobtimus/src/loan.rs
da-kami: I am not exactly sure what the problem is. I think one problem is, that the logic of making the interest rate configurable on top of loan term and collateralization. 

Additionally, is it problematic that I splitted the calculation into small, testable functions?
At the moment there are tests that make sure that each calculation step is sane - but I did not add tests for the complete loan taking (yet). 

I am not sure if it was the best idea to split the calculation into tiny chunks - I can combine more things again and we can write tests that cover bigger pieces of code - would that help or would you just like to have another bigger test on top?

bobtimus/src/loan.rs
bonomat: I feel dumb and don't understand it 😔

bobtimus/src/loan.rs
bonomat: Yes, it could be that the confusing part is how the tests are structured. 

I have a hard time forming the code here into logic without grabbing another piece of paper and writing it out which makes it hard for me personally to see if the logic is sound. 

What I would like to see is something like this (independent of this particular test): 

```rust
// given: 

LTV: 80%
exchange rate start: 10,000 
interest_rate: 5%p.a.
loan duration: 30 days
principal amount: 10,000
overcollateralization_percent: 200%
overcollateralization_btc: 3

// when
do_some_math()

// then
collateralization_thresholds: X,
interest: Y, 
...


```

bobtimus/src/loan.rs
```diff
@@ -304,6 +361,63 @@ mod tests {
...
+            base_interest_rate: dec!(0.05),
+
+            // TODO: Specify terms once term validation is implemented
+            terms: vec![],
+            collateralizations: vec![],
+
+            // irrelevant for this test
+            rate: Rate {
+                ask: Default::default(),
+                bid: Default::default(),
```
bonomat, line 454: It might still be useful to state it here so that one does not need to do the math on how the collateral_amount was chosen. 

As a reference, that's what I came up with: 

```
// collateral_amount / collateralization_percent = value of principal_amount

0.3675 / 1.4 = 0.2625

// so

0.2625 BTC = 10,000

// which means, 

10,000 * 1/0.2625 = BTC price 

rate = 38095 USDT/BTC

```

now with this and the given `max_ltv` of `0.75` I get: 

```
initial_price * LTV = liquidation_threshold
38095 * 0.75 = 28571.25
```

I ignored the 0.05% interest rate but the value seems to be close to what you have here.

bobtimus/src/loan.rs
```diff
@@ -352,6 +358,12 @@ fn validate_loan_is_acceptable(
         }));
     }
 
+    if terms.iter().find(|a| a.days == request_term).is_none() {
```
bonomat, line 408: Clippy is not complaining? 

You could also `.any` here (but not that important): 


```
    if !terms.iter().any(|a| a.days == request_term) {
        return Ok(Err(LoanValidationError::TermNotAllowed {
            term: request_term,
        }));
    }
```

bobtimus/src/loan.rs
```diff
@@ -351,6 +380,24 @@ fn validate_loan_is_acceptable(
...
 
+    // If no collateraliztion thresholds are specified in the offer then we ignore this check and only check for LTV
+    // Note that as a safety net the LTV still outweights the collateralization check.
+    if !collateralizations.is_empty() {
+        let mut sorted_collateralizations = collateralizations.clone();
+        sorted_collateralizations.sort_by(|a, b| a.collateralization.cmp(&b.collateralization));
+        let min_collateralization = sorted_collateralizations
+            .first()
+            .context("Unable to determine minimum collateralization")?
+            .collateralization;
```
bonomat, line 391: You could get rid of this error/context handling here if you'd do : 


```
let mut sorted_collateralizations = collateralizations.clone();
    sorted_collateralizations.sort_by(|a, b| a.collateralization.cmp(&b.collateralization));
    if let Some(Collateralization {
        collateralization, ..
    }) = sorted_collateralizations.first()
...

```

bonomat:
I see one issue (somehow related but not a blocker): in the UI we always show the principal amount instead of the repayment amount. Unfortunately it does not seem to be possible to get hold of this number from what is available in https://github.com/comit-network/waves/blob/06701f852298e28bbba69d5d83ed44e36301775a/extension/wallet/src/wallet/extract_loan.rs#L11

bonomat (approved):
Tests run locally. One issue detected and recorded in https://github.com/comit-network/waves/issues/242

PR #237, opened 2021-08-12
Follow-up of #226

bonomat:


extension/wallet/src/storage.rs
```diff
@@ -10,14 +11,11 @@ pub struct Storage {
...
-            .get_item::<String>("open_loans")
-            .context("no key \"open_loans\" in local storage")?
-        {
-            Some(loans) => serde_json::from_str(&loans)?,
-            None => Vec::<LoanDetails>::new(),
-        };
+    pub fn get_open_loans(&self) -> Result<Vec<LoanDetails>> {
+        let loans = self
+            .get_json_item::<Vec<LoanDetails>>("open_loans")
+            .context("no key 'open_loans' in local storage")?
```
thomaseizinger, line 17: I think that error message is not quite right. The result here represents deserialization errors.

The empty case is handled with the `unwrap_or_default`.

extension/wallet/src/storage.rs
```diff
@@ -31,6 +29,7 @@ impl Storage {
         Ok(storage.into())
     }
 
+    /// return an item parsing it using `from_str`
```
thomaseizinger, line 32: Nit: Rustdoc guidelines suggest a slightly different format: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

extension/wallet/src/wallet/extract_loan.rs
```diff
@@ -86,10 +85,7 @@ pub async fn extract_loan(
     .map_err(Error::LoanDetails)?;
 
     storage
-        .set_item(
-            "borrower_state",
-            serde_json::to_string(&(borrower, loan_details.clone())).map_err(Error::Serialize)?,
-        )
+        .set_json_item("borrower_state", &(borrower, loan_details.clone()))
```
thomaseizinger, line 88: You could avoid this `.clone` if you would change the argument to a reference. Only a reference is needed for serialization.

extension/wallet/src/storage.rs
```diff
@@ -10,14 +11,11 @@ pub struct Storage {
...
-            None => Vec::<LoanDetails>::new(),
-        };
+    pub fn get_open_loans(&self) -> Result<Vec<LoanDetails>> {
+        let loans = self
+            .get_json_item::<Vec<LoanDetails>>("open_loans")
+            .context("no key 'open_loans' in local storage")?
+            .unwrap_or_default();
 
         Ok(loans)
     }
```
thomaseizinger, line 21: A cleaner way of doing this would be to create an extension trait.

```rust
pub trait StorageExt {
	fn get_open_loans(&self) -> Result<Vec<LoanDetails>>;
}

impl StorageExt for Storage {
	
}
```

This way, you can have:

- `Storage` as a domain-independent convenience wrapper around `web_sys::Storage`
- convenience functions on `Storage` for loading / storing specific structs

However, given the new API, I am not sure if this function is even still worth having as it is basically a single expression, so you could even inline it and thereby reduce the indirection.

thomaseizinger (commented):
Nice cleanup!

extension/wallet/src/storage.rs
bonomat: good catch.

extension/wallet/src/storage.rs
bonomat: I opted for removing it all together.

PR #240, opened 2021-08-13
Import wasm blobs directly

thomaseizinger:
If you want to try this PR locally, you need to `yarn link` react-scripts from https://github.com/VasilyShelkov/create-react-extension/pull/32.


thomaseizinger:
This is obsolete as part of #252.

PR #241, opened 2021-08-13
Borrower can select term and collateralization

da-kami:
UI styling and arrangement can be improved.
Needs more information / explanation for the user as well. I would say we hand it over to the itchysats team at this point and they can go on with making it more beautiful :)

image


bonomat:
Counter proposal :) 

image

A few points: 

- I realized interest rate is always on the loan term. It should be p.a., we can then depict how much the user has to repay.
- As you can see in the picture, I would model the collateralization in LTV. I find it easier to understand. 
- additional to the LTV we should show the liquidation rate, i.e. when the user is running in danger of getting liquidated. 


What do you think?

da-kami:
> A few points:
> 
>     * I realized interest rate is always on the loan term. It should be p.a., we can then depict how much the user has to repay.
> 
>     * As you can see in the picture, I would model the collateralization in LTV. I find it easier to understand.
> 
>     * additional to the LTV we should show the liquidation rate, i.e. when the user is running in danger of getting liquidated.
> 
> 
> What do you think?

Thanks for iterating over this! Seeing these changes https://github.com/comit-network/waves/pull/243 make more sense. Will these changes depend on https://github.com/comit-network/waves/pull/243 ? 

The UI looks good and the relationship between the values might help the user's understanding.

I initially left the liquidation rate out because I was not sure we should add it yet. Showing the liquidation rate is tied to the `dynamic liquidation` feature - which we did not tackle yet. There is not oracle yet, Bobtimus actually does not liquidate yet. 
I do like the idea of integrating it in the UI already so we move towards the complete product!

da-kami:
Showing the interest rate per anum is nice, but it becomes clearer that the interest rates for these loans is really high. I think that's why some exchanges don't show it in that way. It looks better to show `1.233%` for 30 days than `15%` per year.

Note: I naively calculate the interest rate for the 30 days as `(15 p.a. / 365) * 30`

---

The interest rate will depend on the loan term, so the values in the dropdown will be dynamic. I think it might be hard to get the UX right using a dropdown for this. If the user had already selected something but then choses a different radio button the dropdown will have to be reset because it is dependent on the term.

PR #243, opened 2021-08-15
Use LTV ratio instead of collateralization

bonomat:
Using LTV ratios instead of collateralization in percent is easier to calculate with due to the simple formula of :
`LTV = repayment_amount / (collateral * rate)`
With this formula alone we can compute if the user provided enough collateral or at what rate the loan will be liquidated given a max LTV ratio.

Signed-off-by: Philipp Hoenisch <philipp@hoenisch.at>

### DIFF
--- a/bobtimus/src/lib.rs
+++ b/bobtimus/src/lib.rs
@@ -44,7 +44,7 @@ pub mod problem;
 pub mod schema;
 
 use crate::loan::{
-    loan_calculation_and_validation, Collateralization, LoanOffer, LoanRequest, Term, ValidatedLoan,
+    loan_calculation_and_validation, LoanOffer, LoanRequest, Ltv, Term, ValidatedLoan,
 };
 pub use amounts::*;
 use elements::bitcoin::PublicKey;
@@ -304,16 +304,7 @@ where
                     interest_mod: Decimal::ZERO,
                 },
             ],
-            collateralizations: vec![
-                Collateralization {
-                    collateralization: dec!(1.5),
-                    interest_mod: Decimal::ZERO,
-                },
-                Collateralization {
-                    collateralization: dec!(2.0),
-                    interest_mod: Decimal::ZERO,
-                },
-            ],
+            initial_ltvs: vec![Ltv { ltv: dec!(0.5) }, Ltv { ltv: dec!(0.7) }],
         }
     }
 
--- a/bobtimus/src/loan.rs
+++ b/bobtimus/src/loan.rs
@@ -6,6 +6,7 @@ use elements::{
     Address,
 };
 use rust_decimal::{prelude::ToPrimitive, Decimal};
+use rust_decimal_macros::dec;
 
 #[derive(Debug, Clone, serde::Serialize)]
 pub struct LoanOffer {
@@ -51,8 +52,8 @@ pub struct LoanOffer {
     /// Interest in relation to terms
     pub terms: Vec<Term>,
 
-    /// Interest rates in relation to collteralization
-    pub collateralizations: Vec<Collateralization>,
+    /// Interest rates in relation to ltv
+    pub initial_ltvs: Vec<Ltv>,
 }
 
 #[derive(Debug, Clone, Copy, serde::Serialize)]
@@ -62,12 +63,11 @@ pub struct Term {
     pub interest_mod: Decimal,
 }
 
-/// Allows to specify a better rate for users that
+/// LTV (Loan-to-value) ratio defines the relation of loan value to collateral value, i.e.
+/// (principal_amount / (collateral_amount * current_rate)
 #[derive(Debug, Clone, Copy, serde::Serialize)]
-pub struct Collateralization {
-    pub collateralization: Decimal,
-    /// Interest to be added on top of the base interest rate for this term.
-    pub interest_mod: Decimal,
+pub struct Ltv {
+    pub ltv: Decimal,
 }
 
 // TODO: Make sure that removing sat_per_vbyte is OK here
@@ -76,7 +76,7 @@ pub struct LoanRequest {
     /// Loan term in days
     pub term: u32,
     pub principal_amount: LiquidUsdt,
-    pub collateralization: Decimal,
+    pub ltv: Decimal,
     pub collateral_amount: LiquidBtc,
     pub collateral_inputs: Vec<Input>,
     pub borrower_pk: PublicKey,
@@ -96,12 +96,15 @@ struct LoanValidationParams {
     request_principal: LiquidUsdt,
     min_principal: LiquidUsdt,
     max_principal: LiquidUsdt,
-    request_ltv: Decimal,
+    /// the LTV ratio the user has actually provided, i.e. it's calculated using the provided collateral
+    provided_ltv: Decimal,
+    /// the max allowed LTV
     max_ltv: Decimal,
     request_term: u32,
     terms: Vec<Term>,
-    request_collateralization: Decimal,
-    collateralizations: Vec<Collateralization>,
+    /// the LTV ratio the user has requested
+    requested_ltv: Decimal,
+    ltvs: Vec<Ltv>,
 }
 
 pub fn loan_calculation_and_validation(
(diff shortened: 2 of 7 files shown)


bobtimus/src/loan.rs
```diff
@@ -334,19 +329,29 @@ enum LoanValidationError {
...
+    #[error(
+        "The requested ltv {requested_ltv} was not offered (available ltvs: {:?})",
+        offered_ltvs
+    )]
+    InvalidRequestedLtv {
+        requested_ltv: Decimal,
+        offered_ltvs: Vec<Decimal>,
+    },
+    #[error(
+        "The provided ltv {provided_ltv} does not match the not selected LTV {requested_ltv})"
```
da-kami, line 350:
```suggestion
        "The provided ltv {provided_ltv} does not match the selected LTV {requested_ltv})"
```

bobtimus/src/loan.rs
```diff
@@ -531,17 +557,18 @@ mod tests {
                 days: 30,
                 interest_mod: Decimal::ZERO,
             }],
-            collateralizations: vec![],
+            // not relevant for test but to show as reference - that's what the borrower can chose
+            initial_ltvs: vec![Ltv { ltv: dec!(0.70) }],
 
             // irrelevant for this test
```
da-kami, line 563: I suppose this comment should be removed given that you fill in values?

bobtimus/src/loan.rs
```diff
@@ -562,7 +589,7 @@ mod tests {
         // The `liquidation_price` is calculated by :
         // `(repayment_amount / current_ltv) / collateral_amount`
         // or with numbers in this example:
-        // (10_500 / 0.75 ) / 1.5 = 9333.33333333
+        // (10000 / 0.75 ) / 1.5 = 9333.33333333
```
da-kami, line 592: Why is the repayment amount `10000` now, if the `principal` and `interest_rate` did not change? 
That cannot be correct - with a `principal` of `10,000` and an interest rate of `5%` the `repayment_amount` should still be `10,500` - or am I missing something?

waves/src/Bobtimus.tsx
```diff
@@ -56,14 +51,14 @@ export interface LoanOffer {
...
-    collateralizations: Collateralization[];
+    initial_ltvs: Ltv[];
 }
 
 export interface LoanRequest {
     principal_amount: number;
     collateral_amount: number;
     collateral_inputs: { txin: OutPoint; original_txout: any; blinding_key: string }[];
-    collateralization: number;
+    ltv: number;
```
da-kami, line 61: Why does the borrower send an LTV? 
The current LTV of the borrower can be calculated from `=REPAYMENT_AMOUNT/(LOAN_REQUEST_COLLTERAL*CURRENT_RATE)`. This feels weird, because the borrower now sends values that are used to calculate the LTV, but at the same time provides an LTV. Is this just to cross check that the calculation on UI and Bobtimus is the same?

e2e_tests/e2e_test_setup.sh
```diff
@@ -12,7 +12,7 @@ while getopts "CS" o; do
             then
                 echo "Clearing docker compose data"
                 cd nigiri
-                docker-compose down
+                docker compose down
```
da-kami, line 15: nit: These changes (if committed on purpose) would be nice in a separate commit. I don't see a relation to the code changes.

bobtimus/src/lib.rs
```diff
@@ -304,16 +304,7 @@ where
...
-                Collateralization {
-                    collateralization: dec!(1.5),
-                    interest_mod: Decimal::ZERO,
-                },
-                Collateralization {
-                    collateralization: dec!(2.0),
-                    interest_mod: Decimal::ZERO,
-                },
-            ],
+            initial_ltvs: vec![Ltv { ltv: dec!(0.5) }, Ltv { ltv: dec!(0.7) }],
```
da-kami, line 307: I am not sure about the the name `initial_ltv` - are these thresholds for the borrower?

bobtimus/src/loan.rs
```diff
@@ -51,8 +52,8 @@ pub struct LoanOffer {
     /// Interest in relation to terms
     pub terms: Vec<Term>,
 
-    /// Interest rates in relation to collteralization
-    pub collateralizations: Vec<Collateralization>,
+    /// Interest rates in relation to ltv
```
da-kami, line 55: The `Ltv` struct does not represent a relation to the interest rate. I think this comment is incorrect.

da-kami:
An overall note on price vs. rate:

In my previous PR I started introducing price when talking about the exchange rate. It seems that this is not helping, because now we started mixing `price` and `rate` in code and comments.

How about we remove the `price` again, but rename the `rate` to `exchange_rate` so it is clear we are talking about the exchange rate and not the interest rate in calculations?

bobtimus/src/loan.rs
```diff
@@ -62,12 +63,11 @@ pub struct Term {
...
 #[derive(Debug, Clone, Copy, serde::Serialize)]
-pub struct Collateralization {
-    pub collateralization: Decimal,
-    /// Interest to be added on top of the base interest rate for this term.
-    pub interest_mod: Decimal,
+pub struct Ltv {
+    pub ltv: Decimal,
 }
 
 // TODO: Make sure that removing sat_per_vbyte is OK here
```
da-kami, line 73: sidenote: this comment is outdated and can be removed (sry, I forgot to remove it)

bobtimus/src/loan.rs
```diff
@@ -110,24 +113,29 @@ pub fn loan_calculation_and_validation(
...
     current_price: LiquidUsdt,
 ) -> Result<ValidatedLoan> {
-    let interest_rate = calculate_interest_rate(
-        loan_request.term,
-        loan_request.collateralization,
-        &loan_offer.terms,
-        &loan_offer.collateralizations,
-        loan_offer.base_interest_rate,
-    )?;
+    let selected_term = match loan_offer
```
da-kami, line 116: This is validated double now - there is still a check in `validate_loan_is_acceptable` for the term. 
My idea was to keep **all** the validation steps in one place and accept that there are some calculations that we might do beforehand even though the loan is not acceptable. 

With introducing this early exit scenario we have to maintain this validation step twice in code. 
Either we remove it here again, or we remove it in the validation step. If we remove the validation step we split up the validation again and don't contain it in one function.

bobtimus/src/lib.rs
da-kami: OK, given the code in `validate_loan_is_acceptable` this is not an interval, but we check against exact matches.

Most user interfaces I looked at don't let the user chose the LTV but the collateralization rate - and then calculate the LTV. I think with this change the user will have to select the LTV. I think we can do that, just pointing out the implication.

bobtimus/src/loan.rs
```diff
@@ -403,27 +408,34 @@ fn validate_loan_is_acceptable(
...
-            .collateralization;
-
-        if request_collateralization < min_collateralization {
-            return Ok(Err(LoanValidationError::CollateralizationBelowMin {
-                request_collateralization,
-                min_collateralization,
-            }));
-        }
+    // we allow for slight price variations of 1% up or down
+    let one_percent = requested_ltv
```
da-kami, line 412: With this we have two price fluctuation intervals now - one for the actual accepted eschange rate and one for the  LTV.

bobtimus/src/loan.rs
```diff
@@ -809,41 +829,79 @@ mod tests {
...
+            .with_provided_ltv(provided_ltv)
+            .with_ltvs(ltvs);
+
+        validate_loan_is_acceptable(loan_validation_params)
+            .unwrap()
+            .unwrap();
+    }
+
+    #[test]
+    fn given_loan_request_with_in_list_is_ok() {
```
da-kami, line 896:
```suggestion
    fn given_loan_request_within_list_is_ok() {
```

bobtimus/src/loan.rs
```diff
@@ -809,41 +829,79 @@ mod tests {
...
+        assert_eq!(
+            error,
+            LoanValidationError::InvalidProvidedLtv {
+                requested_ltv,
+                provided_ltv
+            }
+        );
+    }
+    #[test]
+    fn requested_ltv_is_slightly_different_to_provided_ltv() {
```
da-kami, line 880:
```suggestion
    fn given_requested_ltv_is_slightly_different_to_provided_ltv_then_ok() {
```

bobtimus/src/loan.rs
bonomat: It's still irrelevant but  I want to have the values there nevertheless :)

da-kami:
Other things I noticed that we should clean up before merging:

* I think the `max_ltv` is not needed anymore because this check should already be covered by finding the `requested_ltv` in the `initial_ltvs`

e2e_tests/e2e_test_setup.sh
bonomat: wasn't on purpose...

